PR #38, opened 2015-02-08
Fix for uncaught numeric parse exception

SoujiSeta:
Hello kwsch, I have added a fix for an uncaught system format exception in the SAV_Trainer::change255 function due to the text box containing trailing underscores.


kwsch:
Thanks, nice catch!

PR #40, opened 2015-02-11
Add Level to Box Data Report

Kaisonic:
Added Level to PKX class, and thus Box Data Report. Feature requested here: http://projectpokemon.org/forums/showthread.php?44969-Box-Data-Report-Level


kwsch:
Thanks, and nice improvement of the form :gem:

PR #44, opened 2015-03-05
Fixing a bug with the Country Sub Region selection

FullLifeGames:
This bug occurred via switching from "Australia" to "Austria" while selecting "Vorarlberg".
It throws an ArgumentException and this is the fix I used while working on this: http://projectpokemon.org/forums/showthread.php?44994-Mass-Editor-for-all-PKHeX-supported-files


kwsch:
Thanks for finding this bug and submitting a fix :+1:

PR #56, opened 2015-09-13
Blank space files args and fixing a bug

FullLifeGames:
- Support for files with a blank space in their name by adding the arguments together
- fixing a bug which makes them Abomasnow, by moving the "CB_Species.SelectedIndex = 1;" definition

### DIFF
--- a/PKX/f1-Main.cs
+++ b/PKX/f1-Main.cs
@@ -141,13 +141,22 @@ namespace PKHeX
             pathSDF = Util.GetSDFLocation();
             path3DS = Util.get3DSLocation();
             if (args.Length > 1)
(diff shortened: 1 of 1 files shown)


kwsch:
Breaks if you try to start PKHeX by dragging on two files.

openQuick handles a single path.

kwsch:
Committed a different fix but should be adequate:
https://github.com/kwsch/PKHeX/commit/4cc9546504949952e7deb9539eb3316f09a52012

It'll load the multiple arguments now, and it won't override the species.
Turns out the species setting is done earlier and this one (that was removed) is a really old leftover.

PR #67, opened 2016-01-08
Pokemon feelings not being saved correctly after closing dialog box

chenzw95:
After saving changes in the Memories Editor, the pokemon memory feeling will always be saved as 0. This is because CB_OTFeel and CB_CTFeel are not bound to a datasource in getLangStrings(), and thus SelectedValue will always return null, and Util.GetIndex will always return 0. Fix is to use SelectedIndex instead.


kwsch:
Thanks!

PR #78, opened 2016-03-08
zh translation fix

easyworld:
* update some bad translation in lang_zh.txt
* translate some(not all) ja words and remove some useless characters


kwsch:
Thank you!

PR #92, opened 2016-03-28
Extra checks for GeoLocation legality

OliverG96:
- Make sure latest residence matches current handler 
- If the Pokemon was traded but is actually on OT's savefile that means it should have at least two residences registered

### DIFF
--- a/Legality/Checks.cs
+++ b/Legality/Checks.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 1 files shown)


kwsch:
Won't work as intended if:
- No save file is loaded
- Searching through Database for Legal/Illegal, where individual .pk6 files aren't associated with a save file.

There's also a potential issue with referencing the save file in every legality check (cpu & memory use).

PKHeX's Set to SAV option (Modify PK6 Info) automatically updates the residence/HT info when set to a slot in a save file.

May be best to pass the individual save file parameters depending on the source of the pk6 data.

OliverG96:
I feel so dumb now, I guess I got way to used to load the entire savefile when working on pkhex so I totally miss there's a chance people won't load a entire save everytime

PR #97, opened 2016-04-15
Check for illegal hatched species

chenzw95:
- Ditto
- Most of the Undiscovered egg group

### DIFF
--- a/Legality/Tables.cs
+++ b/Legality/Tables.cs
@@ -444,6 +444,68 @@
             538, 539, // Sawk & Throh
         };
 
+        #region Legality; unhatchable from egg
+
+        internal static readonly int[] NoHatchFromEgg =
+        {
+            132, // Ditto
+            144, // Articuno
+            145, // Zapdos
+            146, // Moltres
+            150, // Mewtwo
+            151, // Mew
+            201, // Unown
+            243, // Raikou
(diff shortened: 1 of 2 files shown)


Legality/Tables.cs
```diff
@@ -444,6 +444,68 @@ public static partial class Legal
...
+            385, // Jirachi
+            386, // Deoxys
+            480, // Uxie
+            481, // Mesprit
+            482, // Azelf
+            483, // Dialga
+            484, // Palkia
+            485, // Heatran
+            486, // Regigigas
+            487, // Giratina
```
gocario: What about Dialga/Palkia/Giratina obtained from Arceus in the HG/SS event?

chenzw95: I went back to my copy of SS to take a look, the level 1 Giratina that I have does not reflect that it hatched from an egg in the summary screen (I can't remember how Gen IV summary works, it should say "hatched" if it came from an egg, right?).

Also, the legality checker does not run for pre-Gen VI Pokemon, so such Pokemon will not be flagged by the legality checker.

suloku:
They hatching from an egg is part of the event/storyline. You never get an egg you can hatch, you receive them as level 1 pokemon.

PR #100, opened 2016-04-19
Fix Legality Check for some Wonder Cards

Wheth:
As commit message says. Lack of return here made some legal event Pokémons (those that match only one Wonder Card, like ORAS Shiny Beldum) show up as illegal.


kwsch:
Ha, nice catch!

PR #128, opened 2016-07-03
Bump to .NET Framework v4.6.1 + Friends

Tsunamical:
As per commit titles and descriptions, these bring nice, small improvements. Builds and functions as expected with Visual Studio, untested on Mono as of the moment but I don't doubt it should function as expected when compiled through Mono as well.


kwsch:
Possibly breaks Windows XP compatibility, as afaik the last .NET Framework available to that OS is 4.0

Tsunamical:
That's correct. There may be hacky ways to do get it working on XP, but I can leave the first commit up in case there's ever a desire to move to that revision. I can separate the second commit into a different PR though, if you would like.

kwsch:
Yeah that'd be great. I'll merge the 'small code improvements' pr; if you submit the other one I'll just close it ;)

PR #136, opened 2016-07-18
Final Set of Code Improvement Patches

Tsunamical:
If anything seems off or questionable, let me know. Built and everything functions as expected, main point of these are to prevent rare NPEs and improve code style.

### DIFF
--- a/MainWindow/Main.cs
+++ b/MainWindow/Main.cs
@@ -92,7 +92,7 @@ namespace PKHeX
             #endregion
             #region Localize & Populate Fields
             string[] args = Environment.GetCommandLineArgs();
-            string filename = args.Length > 0 ? Path.GetFileNameWithoutExtension(args[0]).ToLower() : "";
+            string filename = args.Length > 0 ? Path.GetFileNameWithoutExtension(args[0])?.ToLower() : "";
             HaX = filename.IndexOf("hax", StringComparison.Ordinal) >= 0;
 
             // Try and detect the language
@@ -698,7 +698,7 @@ namespace PKHeX
 
             SAV.FilePath = Path.GetDirectoryName(path);
             SAV.FileName = Path.GetExtension(path) == ".bak"
-                ? Path.GetFileName(path).Split(new[] {" ["}, StringSplitOptions.None)[0]
+                ? Path.GetFileName(path)?.Split(new[] {" ["}, StringSplitOptions.None)[0]
                 : Path.GetFileName(path);
             L_Save.Text = $"SAV{SAV.Generation}: {Path.GetFileNameWithoutExtension(Util.CleanFileName(SAV.BAKName))}"; // more descriptive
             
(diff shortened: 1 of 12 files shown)


MainWindow/Main.cs
```diff
@@ -92,7 +92,7 @@ public Main()
             #endregion
             #region Localize & Populate Fields
             string[] args = Environment.GetCommandLineArgs();
-            string filename = args.Length > 0 ? Path.GetFileNameWithoutExtension(args[0]).ToLower() : "";
+            string filename = args.Length > 0 ? Path.GetFileNameWithoutExtension(args[0])?.ToLower() : "";
```
evandixon: Not needed.  args[0] will always be the path of the executable, so it will never be null.

What's more, the ?. operator will return null, which would cause a build error below on filename.IndexOf.  Because this will never be the case, however, that's not a problem.

MainWindow/Main.cs
```diff
@@ -2470,7 +2470,7 @@ private void clickExportSAV(object sender, EventArgs e)
             if (SAV.HasBox)
                 SAV.CurrentBox = CB_BoxSelect.SelectedIndex;
 
-            bool dsv = Path.GetExtension(main.FileName).ToLower() == ".dsv";
+            bool dsv = Path.GetExtension(main.FileName)?.ToLower() == ".dsv";
```
evandixon: I don't think Path.GetExtension will return null.  If anything, it would return an empty string, which can still have ToLower called

evandixon:
I could be wrong, since I'm much more familiar with VB.Net than C#, but the casting you're using converts from one type to another, while using "as" directly casts.  The general consensus online is that "as" is faster than casting.  The downside is that in the unlikely event the object is not of the target type, a null reference exception is thrown instead of an invalid cast exception, making it a little harder to debug.  I'm not really sure which is preferred in this project: marginal speed improvement (probably not noticeable for relatively small applications like this) or better debugging experience in the event things go wrong.

Sources for performance: 
http://theburningmonk.com/2010/04/net-tips-using-as-vs-casting-in-csharp/
http://buildstarted.com/2014/06/27/casting-in-csharp/

MainWindow/Main.cs
```diff
@@ -785,10 +785,10 @@ private void openSAV(byte[] input, string path)
...
                     getPKMfromFields = preparePK6;
                     extraBytes = PK6.ExtraBytes;
-                    TB_GameSync.Enabled = (SAV as SAV6).GameSyncID != 0;
-                    TB_GameSync.Text = (SAV as SAV6).GameSyncID.ToString("X16");
-                    TB_Secure1.Text = (SAV as SAV6).Secure1.ToString("X16");
-                    TB_Secure2.Text = (SAV as SAV6).Secure2.ToString("X16");
+                    TB_GameSync.Enabled = ((SAV6) SAV).GameSyncID != 0;
+                    TB_GameSync.Text = ((SAV6) SAV).GameSyncID.ToString("X16");
+                    TB_Secure1.Text = ((SAV6) SAV).Secure1.ToString("X16");
+                    TB_Secure2.Text = ((SAV6) SAV).Secure2.ToString("X16");
```
evandixon: Not so much a comment on your PR, but SAV casted to SAV6 should probably be put in a variable so we don't keep casting.

While it's possible that the .Net optimizer would handle this, it might look better.

Tsunamical:
As of .NET 2.0, performance between the two is roughly equivalent, I can't remember the source, but I'll try to find it again. I'll go through and handpick cases that may actually cause NPEs and simplify the multiple casts. Thanks for the suggestions!

kwsch:
Merging, then doing further tweaks/improvements as noted.

PR #168, opened 2016-08-08
Change text_Pokeblock_*.txt include type to None

evandixon:
The text_Pokeblock_\*.txt files are the only text files with a content type of Content, instead of None like all the others.  For Windows Forms applications, I am not aware of any difference this makes.  However, when importing a nuget package into a web application, these files would be added to the project.  Seeing that PKHeX opens these files as embedded resouces, this is not necessary, and adds unnecessary clutter.

I included the icon in these changes for the same reason.  After quickly testing everything out, I notice no side effects.


evandixon:
Closing PR since there's more changes inbound that I might as well include in another PR.

Issue #171, opened 2016-08-09
Met_Year and others not overridden by PK3 + Date suggestion


evandixon:
### The bug
Someone recently uploaded a 3rd gen Bulbasaur to my new PKM DB, and things broke.
http://pkmdb.skyeditor.org/Pokemon/Details/f1ddb01d-ea71-41e5-b851-b743aa06236c

The error was that the met year, month, and day are all 0.  I updated my code to handle that better, but now should be fixed in PKHeX.  I want to say it's a simple as overriding these properties in PKHeX.PK3, but I don't know off-hand how it's stored.  I can take care of this if someone can provide a link to relevant  and thorough enough research.

IMO the majority of the properties in PKHeX.PKM should be abstract instead of virtual (especially properties that aren't relevant in all generations, like Characteristic), but that's not important right now.

### My suggestion
I noticed there's several places where `try { CAL_MetDate.Value = new DateTime(pk5.Met_Year + 2000, pk5.Met_Month, pk5.Met_Day); }` is used (for each generation).  I propose that a property be added to PKHeX.PKM handle this.  Unless there's any disagreements, *I* can take care of this part (tomorrow probably).

### Tasks
- [ ] Hide Met Date combo box for Gen 3 Pokemon (Maybe?)
- [x] Add Met_Date to PKM that returns a date based on Year, Month, and Day, or 1/1/2000 if any component is 0. [Assigned to evandixon]
- [x] Update references to Met_Year and others to the new Met_Date property, where appropriate. [Assigned to evandixon]

As an off-topic aside, would anyone be opposed to me adding some basic unit tests?  They wouldn't be very comprehensive, but I might be able to add some for new code I write.

kwsch:
`PK1`/`PK2`/`PK3`do not store Met Date or Egg Date as there is no datetime for the system. There's no property for this information for these formats.

By using the `virtual` modifier in the parent class, a default can be set so that a consistent implementation is used for all child classes behave the same way.

All properties that fall under this category are marked virtual:
https://github.com/kwsch/PKHeX/blob/master/PKM/PKM.cs#L127-L139

Changing the implementation to return -1 and using that as an indication whether or not that field really exists is a possible solution. **However, you can already use 0 as the indication.** The child classes that actually have this field return >0, >0, >2000 (m/d/y).

May be best to just change the virtual implementations to return -1 as a sign to not have the data, but it falls apart for unsigned numbers (like PID), where they don't exist for PK1/PK2.

[...]

PR #172, opened 2016-08-10
Refactoring: Met Date Property & Tests

evandixon:
This is a seemingly big change, but I don't think it's as bad as it looks.

Also addresses Issue #171
### Moves PKHeX project to its own directory
Instead of having the solution and project in the root, the project is now in a sub-directory.  This makes it easy to add more projects, like unit tests.
### MetDate
The UI can now have a higher-level knowledge of the met date, not having to manually interact with year, month, and date components.  The components are now protected, since it's the class's job to do low-level data access.  Outside the class, there's now a single property of type `DateTime?`.  Just set that to your target date, and this new property will take care of everything.  Setting it to null will zero out the components, previously done in the main forms when a Pokémon is still an egg.

This change affects `PKM`, `PGF`, and `WC6`.
### Tests
I've added tests for my new code, to help ensure things don't break.  Now that there's a test project, it may be easier for others to write new tests too.  (Yes, I know they're annoying to write, but they help prevent bugs and regression.)

I didn't add tests for the new property in `PGF` and `WC6`, since the property does the same thing.  If anyone *wants* a test for that property, I can write one before this PR is completed.


kwsch:
I don't think DateTime will work for specifying a date in the Batch Editor. Will have to double check that the M/D/Y properties can still be modified now that they are `protected`.

PR #173, opened 2016-08-10
Fixed batch editor and database searches on properties of type `DateT…

evandixon:
Fixed batch editor and database searches on properties of type `DateTime?`


PKHeX/Util/ReflectUtil.cs
```diff
@@ -38,5 +38,26 @@ internal static bool HasProperty(this Type type, string name)
...
+                if (DateTime.TryParse(value.ToString(), out dateValue))
+                {
+                    return new DateTime?(dateValue);
+                }
+                else
+                {
+                    return null;
+                }
+            }
+            else
```
evandixon: If there's ever any other types that Convert.ChangeType can't handle, here's where an "else if" for that type should be added.

PR #190, opened 2016-08-21
Move learns data from gen 3, 4 and 5

javierhimura:
I have added files with the move learns by leveling up, eggmoves, hmtm and tutors moves from gen 3, 4 and 5. The only data missed is the special move tutors for the starter pokemons and draco meteor for dragons and GC tutors and moves of purified pokemon. 

I include the project i make to extract the data from de GBA and NDS roms. Is not data generated throught webs like Bulbapedia, is extracted from the games itself.

With this legal comprobations from old gens can be made. For this i think the Legal analysis classes should be refactored into a non-static class to have a Legal6 class, Legal5 class, Legal4 class, Legal3 class and in the future Legal7 class, because static members are not the best to inherit a class. The refactory would be need , even if not making old gen legal analysis, to have in the future gen 6 and 7 analysis.

PKHex project has only been changed to load the files from old gens, but for now they are not used yet.


javierhimura:
PKHex.ExtractData have a reference to the main PKHex project to use some objects previously defined. It can't be compilled as an stand alone project

ExtractData/PKHeX.ExtractData/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+        static void Main(string[] args)
+        {
+            string folder = Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
+            if (Path.GetFileName(folder).ToLower() == "debug")
+                folder = Directory.GetParent(folder).FullName;
+            if (Path.GetFileName(folder).ToLower() == "release")
+                folder = Directory.GetParent(folder).FullName;
+            if (Path.GetFileName(folder).ToLower() == "bin")
+                folder = Directory.GetParent(folder).FullName;
+            if (!Directory.Exists(folder)) return;
```
evandixon: I don't think line 21 is needed.  The working directory is going to exist 9 times out of 10.

If it doesn't, it would be better to throw an exception, or since this is a console app, write to the console.

ExtractData/PKHeX.ExtractData/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Threading.Tasks;
+using System.IO;
+
+namespace PKHeX.ExtractData
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            string folder = Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
```
evandixon: I would recommend using the working directory instead (Environment.CurrentDirectory).  Most of the time it's the same as the assembly's location, but there are some cases where one would want it to be different.

ExtractData/PKHeX.ExtractData/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+            if (Path.GetFileName(folder).ToLower() == "debug")
```
evandixon: Doing these checks complicates things.  It looks like this program is expecting certain files with very specific names to be in very specific places.

I would recommend just checking for certain file or directory names in the working directory, or at the very least, provide documentation as to where the files should go.

ExtractData/PKHeX.ExtractData/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+                folder = Directory.GetParent(folder).FullName;
+            if (!Directory.Exists(folder)) return;
+            if (!Directory.Exists(folder + "\\Logs\\")) Directory.CreateDirectory(folder + "\\Logs\\");
+
+            ExtTMHM3 TMHM3 = new ExtTMHM3();
+            ExtLearnSet3 Learn3Data = new ExtLearnSet3();
+            ExtMoveTutors3 MoveTutors3 = new ExtMoveTutors3();
+            ExtEggMoves3 EggMoves3Data = new ExtEggMoves3();
+            
+            TMHM3.ExtractTMHM(folder + "\\Logs\\", folder + "\\GBA\\", folder + "\\Output\\");
```
evandixon: It's possible that the C# compiler will optimize this, but to improve performance, I would recommend assigning these "folder + ..." expressions to variables since you're referring to them several more times.

ExtractData/PKHeX.ExtractData/ExtTMHM3.cs
```diff
@@ -0,0 +1,376 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace PKHeX.ExtractData
+{
+    class ExtTMHM3 : ExtMoveTutors4
+    {
+        internal override GenerationData GData => Generation3Data.Create();
+
+        static readonly string[] TMMoves_Gen4 =
```
evandixon: I think it would be better to store this list of strings as an embedded resource.  Not only is it easier to manage, but it makes translation a _whole_ lot easier.

Also this variable has no references.

ExtractData/PKHeX.ExtractData/ExtTMHM3.cs
```diff
@@ -0,0 +1,376 @@
...
+        int[] HM_Gen4_DPPt_MoveId = { 15, 19, 57, 70, 432, 249, 127, 431 };
+        int[] HM_Gen4_HGSS_MoveId = { 15, 19, 57, 70, 250, 249, 127, 431 };
+
+        internal override byte[][] ExtractCompatibilityList(int maxspecies, string filename, int offset, int size)
+        {
+            byte[][] Byte_Data = new byte[maxspecies+1][];
+            //Byte_Data[0] = new byte[size];
+            byte[] RawData = File.ReadAllBytes(filename);
+
+            using (var s = new MemoryStream(RawData))
```
evandixon: If you want the data as a stream, why not open a FileStream?

ExtractData/PKHeX.ExtractData/ExtTMHM3.cs
```diff
@@ -0,0 +1,376 @@
...
+            TMHMTutorMoves[] Tutors_Check_Data = TMHMTutorMoves.getArray(ExtractUtils.unpackMini(File.ReadAllBytes(output_folder + "hmtm_g3.pkl"), "g3"));
+            string diff = string.Empty;
+            bool check = ExtractUtils.CheckEquals(E_MovesCompatibility, Tutors_Check_Data, out diff);
+            if (!check)
+                throw new Exception("Check HM TM Moves Gen 3." + diff);
+
+        }
+
+
+        public int[][] InheritanceTMHMMOves(bool[][] FlagsCompatibility, ushort[] MoveList)
```
evandixon: Nitpick: "MOves"

ExtractData/PKHeX.ExtractData/ExtTMHM3.cs
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        internal void TraceTMHMList(string filename, ushort[] MoveList, int TMCount, int HMCount)
+        {
+            string log = string.Empty;
+            string[] all_moves = Util.getMovesList("en");
+
+            for (int i = 0; i < TMCount; i++)
+            {
+                log += $"TM{(i + 1).ToString("00")} Move {MoveList[i]} { all_moves[MoveList[i]]}" + Environment.NewLine;
```
evandixon: Whenever two strings are concatenated in .Net, an entirely new string is created.  Since you're concatenating to the same string repeatedly, this could be a significant performance hit, so you should use a StringBuilder here.

ExtractData/PKHeX.ExtractData/Utils/ExtractUtils.cs
```diff
@@ -0,0 +1,382 @@
...
+                {
+                    Id_SearchMoves[i] = (ushort)all_moves.
+                            Select((move, index) => new { index = index, move = move }).
+                            Where(x => x.move == move_find).
+                            Select(x => x.index).
+                            First();
+                }
+                else
+                {
+                    object a = new object();
```
evandixon: Not needed.

ExtractData/PKHeX.ExtractData/Utils/ExtractUtils.cs
```diff
@@ -0,0 +1,382 @@
...
+            }
+            else
+            {
+                return 0;
+            }
+        }
+
+        private static string[] getStringList(string txt)
+        {
+            //object txt = Properties.Resources.ResourceManager.GetObject("text_" + f + "_" + l); // Fetch File, \n to list.
```
evandixon: Not needed.

ExtractData/PKHeX.ExtractData/Utils/ExtractUtils.cs
```diff
@@ -0,0 +1,382 @@
...
+                for (int i = 0; i < Moves.Length; i++)
+                {
+                    bw.Write(Moves[i].setBits());
+                }
+                return s.ToArray();
+            }
+        }
+
+
+        internal static bool CheckEquals(TMHMTutorMoves[] Moves_Generated, TMHMTutorMoves[] Moves_Readed, out string diff)
```
evandixon: LINQ provides an extension method called SequenceEqual that can be used like this:
`Moves_Generated.SequenceEqual(Moves_Readed);`

javierhimura: Yes, but i want to know exactly were is the first element where the data starting being different, because the da readed and generated should be the same, a different means an error in the process of writing and reading the data. That dont happens in the last version but it happened before it was completed

ExtractData/PKHeX.ExtractData/Utils/ExtractUtils.cs
javierhimura: Sorry, i forgot to delete that function. That was used to find some of the offset of the moves data, most of the code to find the offset was deleted because is not needed. This function is not used anymore.

ExtractData/PKHeX.ExtractData/ExtTMHM3.cs
javierhimura: I didnt thought about performance hit because this program only need to be executed once to extract the data. I did it and the data is already in the PKHex folder, i could just commit the pkl files but i thought no one could trust the data if i dont show how i extrated it from the roms

ExtractData/PKHeX.ExtractData/ExtTMHM3.cs
javierhimura: TMMoves_Gen4 was used to find the move tutors offset in the rom. Most of the code to find offset was deleted after i have the offsets because is not needed anymore.

ExtractData/PKHeX.ExtractData/Program.cs
javierhimura: As i said in my comment below i didnt think about perfromance while doing this because its only purpose was to run once and generate the pkl files.  I used it, generated the files, the code is included just to show the binary files added to PkHex are valid

ExtractData/PKHeX.ExtractData/Program.cs
javierhimura: The program was made to launch throught VS. I know is not the best thing to do but because once completed there is only needs to launch it once i prepared the program to find always the files while in the bin/debug folder without copying all the roms to the bin folder. Once the development has been completed because there is no need to run the program anymore there is no need to change the folder to something more portable.

javierhimura:
Just to clarify the only files that need to be merged are the ones in PKHex project. ExtractData project is only included to proof that i created the pkl files extracting the data from the games and can be trusted and i didnt make it manually with tables from websites

kwsch:
I agree with excluding the ExtractData data-ripping additions; best kept as a separate repository.

Regarding inclusion of the other files:
Having some legality table info for past gen games is good, but it has to be used to be useful. In order to properly utilize past gen tables, legality checking for 3-5 would have to be added. PKHeX's current legality checks are very useful as the common trade/battle traffic revolves around gen6 origin files... past gen origin -- a little. 

Past-gen legality checking is a very sketchy subject, as a bit of information is lost when transferring to newer games. Checking a pk4 from gen3 would be slightly different than checking the un-migrated pk3, and so on for all the various combos 3-6(7). Also would include PIDIV checking, which is not currently implemented. 

The legality checks for Gen6 are likely to be updated regarding transferring to Gen7, and Gen7<->7 checks would be added. That will definitely take some time to develop and polish.

As far as implementing (1-2)3-5 legality checking, it's not something I have any interest in doing, so these files have no use to me. However, someone else may take interest in implementing any of the following Legality Analysis patterns:

- [ ] Gen 1 #pk1 Legality
- [ ] Gen 2 #pk2 Legality
- [ ] Gen 1 #pk2 Legality
- [ ] Gen 3 #pk3 Legality
- [ ] Gen 3 #pk4 Legality
- [ ] Gen 4 #pk4 Legality
- [ ] Gen 3 #pk5 Legality
- [ ] Gen 4 #pk5 Legality
- [ ] Gen 5 #pk5 Legality
- [ ] Gen 3 #pk6 Legality
- [ ] Gen 4 #pk6 Legality
- [ ] Gen 5 #pk6 Legality

tl;dr - can of worms I'm not interested in opening :\

suloku:
Just a note for pk1-pk2: previously it was thought that all possible IV spreads existed, bt it has cureently been "discovered" that due to how the RNG works, possible IV spreads for each pokemon are tyied to map encounter rate (this is a design flaw). Static encounters and in-game trades should be unnafected

This would be specially important with gen1-7 transfer, I doubt nintendo will care to check.

Due to gen 2 eggs and static encounters it should still be possible to have all IV spreads, so the only exception might be ditto, but since it is available in both gen 1 and 2 maybe between gen 1 and 2 all possible combinations are a fact.
The only obvious illegal pokemon might be Mew and Celebo, but is so little information about the events that a full legality check might not be possible.

Pokemon crystal might have some low legality checking because it keeps the time of the day the pokemon was caught in, but not even nintendo bothered to imput propper data in their own events, as the catched data for PNY celebi is weird (can't remember what it was now, I think it had wrong catch level). This data also had catch location (fateful encounter for this events).

javierhimura:
I understand there is no need to include the files without a legal checking.  I want to try to implement the old gen legality checks. As you said many data is lost after transfer, the legal checks with only determine illegal when it can be proven with the data, everithing legal until proven otherwhise.

I think with a proper hierarchy class and using the PKM class instead PK3,PK4 and so ono the legal analysis can be limited to one per generation with a few variations in the class if the PKM is for the class gen or transfered. For example a GetLearnMoves method will obtain the list of valid moves and call the GetLearnMoves of previous gen if the pokemon is transfered. Also the method will use meet level istead of current level when the pokemon is transfered

The encounter analysis its totally diferent if the pokemon is transfered, at leats for gen 3 and 4, because meet data is lost. The only data that can be checked is if the meet level is above the minimun level of the pokemon in the origin game with 3 level tables, one for wild pokemon that can be in any ball, other for pokemon only in pokeball (gift and eggs) and other for safari zone (safaria ball). 

About the PID-IV relations i have myselft a working algorithm written in C# to calculate the RNG method used by the GBA-NDS games except unown in gen 3.

Fortunately all the unavailable data is lost at the first transfer, the encounter chechk for gen 3 is differnt is in gen 3 or not but it should be the same in other gens. I think the same happens with gen 4. Of course that not means in gen 7 a gen 5 pokemon lost some data that exits in gen 6 files

@ suloku Can you give me some source for the IV spreads in gen 1? Im interested in look at that. About Mew events in gen 1 there is only one IV spred used in every event for that pokemon, including the events from back at the game boy are and now in the VC. Celebi IV spread in events is unknown.

suloku:
About gen 1-2 DVs I can't find the original thread, but found this: http://m.youtube.com/watch?v=BcIxMyf8yHY

Mew: most mews had the same iv spread. Don't know about jap mews. Also, NINTEN mew might have had  different IV spread, as well as th obscure spansh distribution (still trying to get another mew from the event to confirm).
Celebi IVs were random (and it is said shiny rstio was increased... At least at PCNY), and I don't think any seed system was used, so it would be impossible to determine legal celebi OT-ID-IV match

kwsch:
Closed as noted in comment section.

If you implement any past-gen legality checks, feel free to open a new pull request with the associated PKHeX code and legality binaries. Binary file creation code is not needed; can always be a separate repository :+1:

PR #191, opened 2016-08-22
Improved Error Handling & Minor Translation Refactor

evandixon:
There was recently [an unhandled exception reported on PP.org](https://projectpokemon.org/forums/showthread.php?49649-Critical-Error), which nobody could have predicted or even have prevented.  But I think it could have been handled a little nicer, instead of having a generic .Net error.

### Added Window to Display Error Messages
image
Notable features:
- Displays the exception details, with a button that copies the text to the user's clipboard
- Displays a context-specific user-friendly error message above it, set by whatever opens the dialog
- At least partially translatable (the friendly error message, which is not technically part of the form, is currently not)

### Added Unhandled Exception Event Handlers
Whenever there's an unhandled exception in the application, the error form will be shown.

### Added Try/Catch statement to save file auto-detection on load
`path = SaveUtil.detectSaveFile();` is now in a try statement, using the above form on error, in case someone has the file corruption problem, linked to above.

### Language
In order to make ErrorWindow translatable, it needs to know what language the form is set to.  To accomplish this, whenever the language is set in the MainWindow, the culture of the current thread is changed, which the ErrorWindow uses during translation.

I also added a French translation for the ErrorWindow.  Disclaimer: I'm still a French student, so the wording might be a little off.  I'd appreciate it if someone more familiar with French could review it (I'll add a line note marking my translation to make it easier to find).

## Still Needed
I'm not sure what the best way to make the user-friendly messages translatable is.  It seems that the current translation engine is designed to translate forms, not provide specific strings to the code.  This isn't limited just to my new inline strings.  Other things like "Illegal mode activated." and "Decrypted PKM File" in the SaveFileDialog filter are affected too.

In Sky Editor, I use resource files to store strings, so in the code, I use `My.Resources.Language.Whatever` to get the localized "Whatever" string in Language.resx, and there's similar syntax in WPF.  I don't know what the best way to do this in PKHeX would be.  While I want to say resource files, that would be a big change.  However, this is probably a discussion for another day.  Once this PR is completed (and people get a chance to review my code), I'll open an issue for further discussion.


PKHeX/Resources/text/fr/lang_fr.txt
```diff
@@ -361,4 +361,14 @@ L_Country = Pays
...
+! End
+! -----------------------------------------------------
+- DO NOT CHANGE THIS SECTION.
+! ErrorWindow = Erreur
+- Change stuff below this line, not above.
+-------------------------------------------------------
+L_ProvideInfo = S'il vous plaît fournir ces informations lorsque vous rapportez cette error: 
+B_CopyToClipboard = Copier dans les Notes
+B_Continue = Continuer
+B_Abort = Abandonner
```
evandixon: I'd appreciate it if someone more familiar with French could proof-read my translation.

kwsch:
Will the stack trace still provide line numbers if the user does not have the .pdb?

Every release is compiled in Debug mode iirc, so if the linenums work it'd definitely be beneficial -- even though we both knew what caused the above error for the user it'd be easier for the user to figure out on their own!

PKHeX/Resources/text/es/lang_es.txt
```diff
@@ -799,3 +799,13 @@ B_Output = Exportar
 B_Cancel = Cancelar
 B_Save = Guardar
 ! End
+! -----------------------------------------------------
+- DO NOT CHANGE THIS SECTION.
+! ErrorWindow = Error
+- Change stuff below this line, not above.
+-------------------------------------------------------
+L_ProvideInfo = Please provide this information when reporting this error:
```
vacero: I can provide the Spanish translation for this. I'll make a PR when this gets merged, or here it is if you want to include it in this one:

L_ProvideInfo = Por favor, proporciona esta información cuando reportes este fallo:
B_CopyToClipboard = Copiar al portapapeles
B_Continue = Continuar
B_Abort = Abortar

PKHeX/Resources/text/fr/lang_fr.txt
TheSonAlsoRises: >S'il vous plaît fournir ces informations lorsque vous rapportez cette error: 

Veuillez fournir les informations suivantes dans votre rapport d'erreur :

>Copier dans les Notes

Copier dans le presse-papier

PKHeX/Resources/text/es/lang_es.txt
evandixon: Thanks!
I went ahead and added it, since there's already an active PR and all.

evandixon:
@kwsch It looks like the PDB _is_ needed to show filenames and line numbers.  Release mode shouldn't make much difference, but the line numbers might be a little off due to the C# optimizer (which doesn't run in Debug mode).

I also added details about what assemblies are loaded, like with the regular error handler.

Debug With PDB:
image

Debug Without PDB:
image

Release With PDB:
image

Release Without PDB:
image

evandixon:
Update: I added an overload to Util.Error that shows this new error window, and updated appropriate references to the old one to use the new one.  Naturally, some errors without exceptions (like the incorrect size of a save file) still use the old one.

evandixon:
Update:
- Added more information to error details.  In the case of opening files, this contains the path.  See here for example output: http://pastebin.com/jhJSJz9h
- Removed resizing error window
image
- The error message when calling openQuick now leaves open the possibility of other IO errors, like the file path being too long, or the file being deleted before being opened (which I did while in Break mode to get these examples)

Issue #196, opened 2016-08-24
Unhandled Exception Errors When Pressing Event Flags on ORAS


ricksee:
Every time I try to use Event Flags on AS I get an Unhanded Exception Error.  

See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.ArgumentOutOfRangeException: Value of '65285' is not valid for 'Value'. 'Value' should be between 'Minimum' and 'Maximum'.
Parameter name: Value
   at System.Windows.Forms.NumericUpDown.set_Value(Decimal value)
   at PKHeX.SAV_EventFlags.addConstList(String[] list)
   at PKHeX.SAV_EventFlags..ctor()
   at PKHeX.Main.B_OpenEventFlags_Click(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.6.1586.0 built by: NETFXREL2
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework/v4.0.30319/mscorlib.dll
----------------------------------------
PKHeX
    Assembly Version: 1.0.0.0
    Win32 Version: 1.0.0.0
    CodeBase: file:///C:/Users/Marlon/Documents/PKHeX/PKHeX.exe
----------------------------------------
System.Core
    Assembly Version: 4.0.0.0
    Win32 Version: 4.6.1586.0 built by: NETFXREL2
[...]

PR #198, opened 2016-08-25
Changed min and max of NumericUpDown from short to ushort

evandixon:
var mtb = new NumericUpDown
                {
                    Maximum = short.MaxValue,
                    Minimum = short.MinValue,
                    Value = Constants[num[i]],

In the above code snippet, Constants[num[i]] is a ushort, while the min and max of the NumericUpDown are being set to the min and max of short.  This commit fixes that, and by extension, #196.

I'm opening a PR instead of directly pushing both because that's what I'm in the habit of doing now, but also to give others a chance to decide whether or not Constants[num[i]] should really be a short.  I chose to make the range a ushort since that's the smaller/safer of the two possible solutions I see.

### DIFF
--- a/PKHeX/Subforms/Save Editors/SAV_EventFlags.cs
+++ b/PKHeX/Subforms/Save Editors/SAV_EventFlags.cs
@@ -171,8 +171,8 @@ namespace PKHeX
                 };
                 var mtb = new NumericUpDown
                 {
-                    Maximum = short.MaxValue,
-                    Minimum = short.MinValue,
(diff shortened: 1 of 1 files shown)


kwsch:
Error is probably thrown because it's read as a ushort but displayed as a short; 0xFFFF = -1 but is beyond the range.

I'm thinking that the constants can be treated as ushort or shorts as ash can go 65535 iirc, so it'd be best to just show as unsigned.

PR #211, opened 2016-08-30
Add support for editing Gen I files

SciresM:
Features mostly lifted wholesale from Rhydon, but then Rhydon lifted most things wholesale from PKHeX, so really, isn't that how it should be?

Gen II support to come later.


PKHeX/MainWindow/Main.cs
```diff
@@ -840,11 +842,77 @@ private void loadSAV(SaveFile sav, string path)
...
+            SlotPictureBoxes[43].Visible = SAV.Generation > 1;
+
+            CHK_IsEgg.Visible = Label_Gender.Visible = SAV.Generation > 1;
+
+            Label_OTGender.Visible = SAV.Generation > 1;
+
+            if (1 <= sav.Generation && sav.Generation <= 2)
+            {
+                Label_SPD.Visible = TB_SPDEV.Visible = TB_SPDIV.Visible = Stat_SPD.Visible = false;
+                Label_SPA.Text = "Spc";
```
kwsch: Changing language breaks this, recommend adding a new Label that toggles visibility.

Not sure on the positioning; for different operating system scaling modes the defined point values may cause controls to be sized incorrectly... had this issue back when PKHeX had a toggle for box interface. Maybe add extra maskedtextboxes that are also not visible.

I'll probably put all the controls into a TableLayoutPanel that'll handle the re-positioning.

PKHeX/MainWindow/Main.cs
```diff
@@ -1644,30 +1727,31 @@ private void updateRandomIVs(object sender, EventArgs e)
...
-                TB_DEFIV.Text = (Util.rnd32() & 0x1F).ToString();
-                TB_SPAIV.Text = (Util.rnd32() & 0x1F).ToString();
-                TB_SPDIV.Text = (Util.rnd32() & 0x1F).ToString();
-                TB_SPEIV.Text = (Util.rnd32() & 0x1F).ToString();
+                TB_HPIV.Text = (Util.rnd32() % (SAV.MaxIV + 1)).ToString();
+                TB_ATKIV.Text = (Util.rnd32() % (SAV.MaxIV + 1)).ToString();
+                TB_DEFIV.Text = (Util.rnd32() % (SAV.MaxIV + 1)).ToString();
+                TB_SPAIV.Text = (Util.rnd32() % (SAV.MaxIV + 1)).ToString();
+                TB_SPDIV.Text = (Util.rnd32() % (SAV.MaxIV + 1)).ToString();
+                TB_SPEIV.Text = (Util.rnd32() % (SAV.MaxIV + 1)).ToString();
```
kwsch: SAV.MaxIV can still behave like a mask; replace with & SAV.MaxIV?

31 = 11111
15 = 01111

PKHeX/PKM/PKX.cs
```diff
@@ -1617,5 +1634,545 @@ public static ushort getG4Item(ushort g3val)
...
+            {0xB2, "s"},
+            {0xB3, "t"},
+            {0xB4, "u"},
+            {0xB5, "v"},
+            {0xB6, "w"},
+            {0xB7, "x"},
+            {0xB8, "y"},
+            {0xB9, "z"},
+            {0xE1, "{"}, /* Pk */
+			{0xE2, "}"}, /* Mn */
```
kwsch: Tabs vs Spaces

PKHeX/Saves/SAV1.cs
```diff
@@ -0,0 +1,522 @@
...
+                Options = (byte)((Options & 0xF8) | new_speed);
+            }
+        }
+
+        public override uint Money
+        {
+            get { return uint.Parse((Util.SwapEndianness(BitConverter.ToUInt32(Data, Japanese ? 0x25EE : 0x25F3)) >> 8).ToString("X6")); }
+            set
+            {
+                BitConverter.GetBytes(Util.SwapEndianness(uint.Parse(value.ToString("000000"), NumberStyles.HexNumber))).Skip(1).ToArray().CopyTo(Data, Japanese ? 0x25EE : 0x25F3);
```
kwsch: Converting to bytes can just use BitConverter.GetBytes((uint)value) to get byte[4]; no need to use ToString to parse/format a uint -> string -> uint.

SciresM: The ToString actually is necessary -- Money and Coins are stored as Binary Coded decimal -- that is, 0x123456 == $123456.

String conversion seems to me the best way to handle that?

kwsch: Yeah, seems simple enough. 

A comment in the src would be great; the game stores this as a Binary Coded decimal' would prevent confusion for anyone using this code centuries from now :)

Issue #199, opened 2016-08-25
Adding in Check for Updates/New Update Notification/Automatic Updates


ricksee:
Is it possible in the future to allow an option to check for updates of this program to see if they are available and then allow the program to download the update or enter the link to download it?

evandixon:
I've been toying with the idea of deploying PKHeX via ClickOnce.  There was actually [some code that would be useful](https://github.com/kwsch/PKHeX/pull/156) if/when this happens in the future, but I removed it because [it broke compatibility with Mono](https://github.com/kwsch/PKHeX/issues/161).

I might be able to get around this by creating another project that wraps PKHeX, and deploying that one via ClickOnce.  However, last time I tried having a ClickOnce deployment with multiple executables, there were some manifest errors.  Since that was years ago, I'll continue to play around with things to see how this would work.

@kwsch 
If you want to pursue this, I would like for us to do the following for each official release:
- Instead of, or in addition to dates, assign a real version number to each future release, preferably with 3 parts (like 1.0.0).  (My "super-secret" dev ClickOnce deployment currently uses 1.0.0.X, where X is the auto-generated number.)   Since we're already on 1.0.0, I'll need you to pick an arbitrary version to start with.  After that, I recommend using the standard Major.Minor.Patch for releases after that (Major and Minor for new features, Patch for bug fixes).
- Tag releases using the version number.  This would also show source code (plus any binaries we upload) on [Github Releases](https://github.com/kwsch/PKHeX/releases).

kwsch:
Simplest way to not break Mono support is to query a webpage to retrieve the latest version info, then popping open a webpage to download the latest version.

Investigating how PokeGen did it, the program queried a [url on projectpokemon](http://pokegen.projectpokemon.org/vp.txt), if it differed it grabbed /PokeGen.zip, with other detections for lang_%s.txt. Pretty sure that's what the extra version info at the end is for (auto update language translations).

It's something I'd want to get done before S&M, as that's prime to catch the users that are updating for that support. Probably best to just do a simple version# check from a URL, then pop open a webpage after prompting the user.

SciresM:
Any reason not to use the github repo (version.txt resource?) (wiki page) as a URL to query?

evandixon:
I think that managing updates ourselves would require too much re-inventing of the wheel.  PKHeX would need its own update checks, its own form, its own code for replacing the executable...  ClickOnce does all this for us.

[...]

PR #212, opened 2016-08-30
Added project configuration for ClickOnce deployments

evandixon:
When building with ClickOnce-Debug or ClickOnce-Release, `System.Deployment` is included as a reference and checked by `Util.IsClickonceDeployed`.  I've confirmed that when in the existing Debug or Release configurations, System.Deployment is not included in the build, in which case that function just returns false, so building with Mono should™ work fine.

Part of issue #199


evandixon:
The newly added continuous-integration/teamcity-mono status check currently builds PKHeX.csproj in the default configuration (Debug x86), which does not include System.Deployment.

Issue #325, opened 2016-10-18
Possible Option: Adding A Warning About Changes Not Saved


ricksee:
Is it possible to add a warning popup messenge that before you exit the program if the save file wasn't exported or saved that you can add "Are you sure you don't want to save this?" or something like that?

kwsch:
Resolved in #327, thanks @evandixon !

May have to make this an option later if users complain about nuisance.

PR #327, opened 2016-10-18
Added warning when closing PKHeX with unsaved changes

evandixon:
Adds feature described in issue #325.

This PR adds a dialog asking for confirmation if someone tries to close the form when there's unsaved changes.  Saving through `clickExportSAV` marks the current save as no longer modified, and will not display the dialog.  I'm unsure if saving the backup with `clickExportSAVBAK` should mark the save as not modified, but it looks fine as-is.

### DIFF
--- a/PKHeX/MainWindow/Main.Designer.cs
+++ b/PKHeX/MainWindow/Main.Designer.cs
@@ -5647,6 +5647,7 @@
             this.Name = "Main";
             this.StartPosition = System.Windows.Forms.FormStartPosition.CenterScreen;
             this.Text = "PKHeX";
+            this.FormClosing += new System.Windows.Forms.FormClosingEventHandler(this.Main_FormClosing);
             this.DragDrop += new System.Windows.Forms.DragEventHandler(this.tabMain_DragDrop);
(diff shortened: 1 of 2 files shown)


PKHeX/MainWindow/Main.resx
```diff
@@ -129,15 +129,6 @@
   <metadata name="Label_IsShiny.Locked" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
     <value>True</value>
   </metadata>
-  <metadata name="Label_PID.Locked" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
```
evandixon: Not sure what all this is.  I didn't intentionally remove it.

kwsch:
SAV.Edited is referenced by the 'Checksum Info' button; I think that's it.

Should be possible to use Util.Prompt instead of the messagebox.

evandixon:
Unintended changes appear to be resolved

PR #333, opened 2016-10-19
B/detect save null reference

evandixon:
Fixes the issue described here: https://projectpokemon.org/forums/showthread.php?50334-Newer-PKHex-Releases-won-t-run&p=221387&viewfull=1#post221387

Also fixes two other issues I found during testing: the abort button in the Error Window doing nothing, and the form close warning showing when the blank save is loaded.

### DIFF
--- a/PKHeX/Saves/SaveUtil.cs
+++ b/PKHeX/Saves/SaveUtil.cs
@@ -509,6 +509,24 @@ namespace PKHeX
         }
 
         /// <summary>
+        /// Determines if the given data is a valid save file with a valid checksum.
+        /// </summary>
+        /// <param name="data">Save data to check</param>
+        /// <returns>A boolean indicating whether or not the given data is a save file with valid checksums.</returns>
+        public static bool isVariantSAVValid(byte[] data)
+        {
+            var save = getVariantSAV(data);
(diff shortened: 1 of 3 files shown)


PKHeX/Saves/SaveUtil.cs
```diff
@@ -544,7 +562,7 @@ public static string detectSaveFile()
                 possiblePaths.AddRange(getSavesFromFolder(Path.Combine(pathCache), false));
 
             // return newest save file path that is valid (oh man)
-            return possiblePaths.OrderByDescending(f => new FileInfo(f).LastWriteTime).FirstOrDefault(p => getVariantSAV(File.ReadAllBytes(p)).ChecksumsValid);
+            return possiblePaths.OrderByDescending(f => new FileInfo(f).LastWriteTime).FirstOrDefault(p => isVariantSAVValid(File.ReadAllBytes(p)));
```
kwsch: Should be:
`.FirstOrDefault(p => getVariantSAV(File.ReadAllBytes(p))?.ChecksumsValid ?? false);`
Removes the need for the extra method. Basically just null check the save in the linq instead of a separate method.

PR #337, opened 2016-10-29
Minor spanish update for Pokemon Sun and Moon

ajtudela:


kwsch:
The SM stuff is placeholder for now - no point if no public save editing and the demo files are partially scrubbed.

Issue #343, opened 2016-11-09
Garbled Chinese name for Pokemon


samyex6:
The text file `Resources/text/zh/text_Species_zh.txt` is garbled.

kwsch:
That's the way it should be, as far as I know. The new gen7 font hasn't been added; I'm pretty sure they're reserving a portion of the characters for the Pokémon names.

ericjwg:
well.... load separate nickname and species text file should help.
one plain text file(Chinese characters) for displaying the species.
the other one is the game extraction for nickname,

evandixon:
Looks like PR #360 solves this.

Issue #346, opened 2016-11-09
2016.11.10 Update Fixed：text_Species_zh.txt


q67732897:
2016.11.10 Update
Fixed several missing Pokemon
Now it is complete with 803（CHS）
[text_Species_zh.txt](https://github.com/kwsch/PKHeX/files/582372/text_Species_zh.txt)
![3](https://cloud.githubusercontent.com/assets/6880568/20162948/99eb7d76-a733-11e6-86fa-d28af32850f9.jpg)

q67732897:
[text_Species_zh.txt](https://github.com/kwsch/PKHeX/files/582368/text_Species_zh.txt)
2016.11.10

evandixon:
Just to confirm, does this issue still need to be open after PR #360?  (Since I don't know the language, I can't double-check this.)

PR #348, opened 2016-11-09
Fixed ZH species list

evandixon:
Uses the file from issue #346. Should also resolve issue #343.

Credit to q67732897


kwsch:
This appears like a [fan translation](https://github.com/kwsch/PKHeX/pull/348/files#diff-9e626cc68d8f96d71f03982b1b72b30aR733) and not the list used by the games.

easyworld:
The species texts in rom are not regular UTF-8, and it's a little hard to get the mapping.   
I'm trying to get 'official' Chinese species one by one from the game. I'll pull the request later.

evandixon:
Looks like this PR is now redundant because of PR #360

PR #385, opened 2016-11-12
B/memecrypto winxp

evandixon:
# The Problem
* SHA256CryptoServiceProvider and friends are not supported on XP.
[Original Report](http://gbatemp.net/threads/can-someone-complile-the-new-gen-7-pkhex.447884/page-13#post-6825115)

# The Solution
The simple solution is to use SHA256Managed and friends instead.  This, however, undoes the fix introduced in PR #367.  Additions have been made to do the following:
1. Attempt to use SHA256CryptoServiceProvider and friends
2. If a PlatformNotSupportedException is thrown, use SHA256Managed and friends.
3. If an InvalidOperation exception is thrown, this likely means FIPS mode is enabled on Windows XP.  I'm not aware of anything that can be done, so the following actions are taken:
  - Upon opening a SAV7, an error is shown warning the user that saving cannot be done.  Care has been taken to ensure that this isn't shown when loading the default save (i.e. loading the form).
  - Upon attempting to save, a MiscSaveCheck is added warning that the checksum will be broken until saved on an OS without FIPS enabled or on a new OS.

Tested on a WinXP VM, before and after enabling FIPS mode.

### DIFF
--- a/PKHeX/MainWindow/Main.cs
+++ b/PKHeX/MainWindow/Main.cs
@@ -783,11 +783,14 @@ namespace PKHeX
             if (sav == null || sav.Version == GameVersion.Invalid)
             { Util.Error("Invalid save file loaded. Aborting.", path); return; }
 
-            if (sav.RequiresMemeCrypto && !MemeCrypto.CanUseMemeCrypto())
+            if (!string.IsNullOrEmpty(path)) // If path is null, this is the default save
             {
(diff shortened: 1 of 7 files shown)


Tsunamical:
It looks like this is the best solution unfortunately. 

An alternative to this would be allowing an option to change the registry key through .NET to the key provided [here](http://stackoverflow.com/a/3341727) to allow usage of SHA256CryptoServiceProvider and friends using Registry.SetValue if it's even possible, but that also may raise security concerns for people, and it's probably best to *not* mess with Registry. May be worth linking in the forum though, but FIPS really shouldn't be enabled on the typical home computer anyway.

PR #408, opened 2016-11-15
F/improve batch editor

evandixon:
# Overview
When using the batch editor or searching in the database, are you tired of having to just know the IDs of things like moves and items?  Now you don't have to!  Just enter something like `.Ball=Ultra Ball`, and it'll work!

Things that this works on:
* Species
* Natures
* Abilities
* Ball
* Moves
* Relearn moves

# Technical Details
## BatchEditAttribute
When any kind of BatchEditAttribute is on a property, ReflectUtil will defer its usual GetValue, SetValue, and GetValueEquals functionality to the attribute.  The attribute can be anywhere in the inheritance line, but there can only be one.  I don't see any circumstance where there may be more than one, but if there is, the `SingleOrDefault` calls can be replaced with `FirstOrDefault`, and the attribute lowest in the inheritance line will be used.

Attributes inheriting from this class will likely only be able to support certain types of properties.  To help enforce this, there is a SupportedTypes readonly property on the attribute.  If for some reason this attribute is on an unsupported property, an InvalidTypeCast exception can be thrown by the child class.  I've written a unit test that ensures there's no property anywhere in the assembly with an unsupported attribute.
## StringListEnumAttribute
Many properties are conceptually an enumeration, with the logical value being a string in the resources.  Any property with the StringListEnumAttribute (inherits from BatchEditAttribute) will be treated as such for reflection.  Getting its value returns the string.  Setting the value can be done either with an integer as usual, a string containing that integer, or the value in the resource list.  A `EnumItemNotInListException` will be thrown if the target string is not in that list, and the Batch Editor has been updated to recognize this as bad input (reporting bad input instead of an internal error).
## Unit Tests
The following things are tested:
* Reflecting to ensure all attributes in PKHeX.exe are on supported types
* Getting a StringListEnum value
* Setting a StringListEnum using an integer, integer in a string, and string value
* Determining if a StringListEnum value is equal to another value, using both integers and strings (looking back, it looks like I left out integer in a string)
* Ensuring the proper exceptions are thrown, as described above (InvalidCastException and EnumItemNotInListException).


kwsch:
Told ya I could do it in 15 loc max -- ignoring braces :)
https://github.com/kwsch/PKHeX/commit/b438650e5d2c9c8be3ecd0fbf8e03546a1389698

PR #517, opened 2016-11-28
Update Images

NWPlayer123:
Ran through FileOptimizer
(http://nikkhokkho.sourceforge.net/static.php?page=FileOptimizer)

Also compiled on latest Visual Studio Community (Release, x86) to make sure it boots


evandixon:
Comparing builds made by my Team City server, the latest `master` build yields a 6.45 MB executable while this PR yields a 6.36 MB executable.  IMO, the change is insignificant, but comparing the first few file differences in GitHub, I see no reason to reject the PR.

NWPlayer123:
It's more of an optimization for anyone cloning the repo/building from source (e.g. fragmentation), this took img from 1.80 MB to 1.24 MB (1,309,133 bytes), the biggest part is the rest of Resources (looking at the exported project from dotPeek), and I'm not sure how much byte and text can be optimized. It also converts all images into BASE64 which doesn't help either.

PR #541, opened 2016-12-01
Make CRC16 method more readable

RainThunder:


kwsch:
Are you sure that this version produces the same checksum value?

kwsch:
Guess we never checked to see if the compiler-optimized version was really just a simple implementation 👍

PR #569, opened 2016-12-03
Fix geolocation checks

Eskuero:
Trading Pokemon still sets Geolocation so we expect it as long as an Handling Trainer exists.
Also having a HT but not a location is wrong as long as its not an event.
I only checked the first country slot in last case because its the first one written if the Geolocation list is empty, so having Geo2_Country but not Geo1_Country should be flagged as well. GeoX_Region may be empty depending on the player.

### DIFF
--- a/PKHeX/Legality/Checks.cs
+++ b/PKHeX/Legality/Checks.cs
@@ -1080,8 +1080,10 @@ namespace PKHeX
                     pkm.Geo1_Country, pkm.Geo2_Country, pkm.Geo3_Country, pkm.Geo4_Country, pkm.Geo5_Country,
                     pkm.Geo1_Region, pkm.Geo2_Region, pkm.Geo3_Region, pkm.Geo4_Region, pkm.Geo5_Region,
                 };
-                if (geo.Any(d => d != 0))
+                if (geo.Any(d => d != 0) && pkm.HT_Name.Length == 0)
                     return new CheckResult(Severity.Invalid, "Geolocation Memories should not be present.", CheckIdentifier.History);
+                if (pkm.Geo1_Country == 0 && pkm.HT_Name.Length != 0 && !pkm.WasEvent)
+                    return new CheckResult(Severity.Invalid, "Geolocation Memories should be present.", CheckIdentifier.History);
                 
                 if (pkm.XY && pkm.CNTs.Any(stat => stat > 0))
                     return new CheckResult(Severity.Invalid, "Untraded -- Contest stats on SM origin should be zero.", CheckIdentifier.History);


SnowPhoenix:
Can you provide more details about your test conditions under which you found geolocation data on traded Gen 7 Pokémon still present? The geolocation bytes (0x94-0x9D) containing data for Gen 7 files is inconsistent with what I and a few others have found.

To provide more detail on our findings, I was examining an ek7 of a Pokémon that a friend had received in a trade from someone we both know. The ek7 had been extracted from RAM, and I noticed the Residence tab in the Memory/Amie Editor form was empty, despite the fact that the Pokémon had been traded before.

I initially tried exporting the ek7 as a pk7 from PKHeX to look at the individual bytes, and noticed the data showed up afterwards for the converted pk7, both in a hex editor and PKHeX. I then used a CLI version of the [save decryption code we use in Porybox](https://github.com/porygonco/porybox/blob/master/client/add/parseSaveFile.js#L107) to convert the ek7 to pk7 directly. This time, while the rest of the data remained the same, the geolocation bytes from 0x94-0x9D were all zero. As such, I reported this to @kwsch. (To clarify, other NotOT data such as 0x78-0x93 and 0xA2 were present.)

I have just rechecked our findings by doing a few Wonder Trades and extracting the Pokémon received from my exported save file. Bytes 0x94-0x9D are still empty for everything I received.

Eskuero:
I didn't messed up with the Hex editor. All I did was to legally complete Alolan Pokedex. I traded several times on GTS and with friends to fill pages of Pokemon that require a trade evolution. When I finished I went to PkHeX to try and catch some false positives to fix (only got an error on Froslass, which is already fixed).
All of the Pokemon I legally evolved with a trade evolution had Country set in GeoLocation

SnowPhoenix:
I didn't claim you messed up with PKHeX/a hex editor. (I'm assuming you mean PKHeX, though "the Hex editor" is a little ambiguous when I was referring to using actual hex editors such as [HxD](https://mh-nexus.de/en/hxd/) and/or [xvi32](http://www.chmaas.handshake.de/delphi/freeware/xvi32/xvi32.htm) to view raw bytes.)

However, if you did use a build of PKHeX made before 18561259d8eb0b3ce7e572f2e3d2c8992dde11a8 (yesterday), there were a number of things that could've caused it to set data that should not be there (through no fault of your own). I don't even know if I have all of them, but the examples I'm aware of where data could've been incorrectly added include

* exporting the Pokémon from the save to a separate *.pk7
* right-clicking and hitting "Set"
* dragging and dropping the Pokémon (even to the same slot)

Assuming you never wrote any saves modified in PKHeX back to your game, what happens if you extract a fresh copy of your save and use [a more recent build](http://teamcity.skyeditor.org/viewLog.html?buildId=lastSuccessful&buildTypeId=PKHeX_PKHeXBuildPr&tab=artifacts) to view the data?

Eskuero:
Yeah with Hex editor I specifically meant HxD. I've spent the last hours doing a GTS giveaway and all the Pokemon I got had no geolocation memories so yeah it's probably something from an old PkHeX build that messed my data.
So I'm closing this since is wrong ^^

Eskuero:
Closed

PR #580, opened 2016-12-05
Edge case for Z-Ring item count set in 2 in an unedited save

Tailmon:
Well, this is a weird case that I have encountered with the items, I exported my save  and I opened it for the first time with the last build of PKHeX and after that I opened the item editor and checked that the item count for the Z-Ring was 2 instead of 1, if I simply click on save without editing anything the item count for the Z-Ring is set to 1, I don't know if the count of Z-Ring is intended to be 2 instead of 1, and this seems strange.
Btw, The original save was never edited or opened with any tool, so it is a normal play through. And also I don't know if this case is repeated in other games.
The status of the play is after defeating the Elite four for the first time
I also attaching the original save file and a modified file that results if I repeat the steps that I mentioned earlier.
https://drive.google.com/file/d/0ByljblyKtoFmbnk5Z2w4VEJJUFU/view?usp=sharing
Greetings!


iSharingan:
I believe the second iteration to the Z ring is from when you get the Key Stone (that's about when the duplicate showed up in my save). It doesn't affect gameplay to set it back to 1 and having 2 doesn't add or remove any functionality in-game.

Issue #544, opened 2016-12-01
Battle Box support in SM


NinFanBoyFTW:
Hey I wanted to know if there will be support for the battle boxes in the pc for sun and moon. Im sure your probably already aware but in the pc you can go to the battle box section and now there is 5 teams you can use instead of the previous one team. So I wanted to know if you were going to add support to add pokemon to the battle box teams to make it easier to set up teams

EDIT: It seems that when you "register" a pokemon to the battle box it stays in the pc with a battle box symbol on it so I guess if you were to add it would would have to be read only which I think it usually is. Thanks

kwsch:
It's a bit much to do anything about. Basically the game stores which box/slot is used for a given team.

PKHeX allows set/delete in addition to box filling/deleting/cloning. 
How would PKHeX handle slots that are used by the Battle Box?
How would PKHeX handle slots that are overwritten by many of the 'cheat' operations listed above?
How would PKHeX handle moving/swapping a pkm between party/slot?

PKHeX isn't a box simulator, and obviously there are tons of what ifs. The easiest solution is to do nothing about it, because players can use their game to manage their battle box.

Plus, modifying battle box contents is NOT something I'd want to do, considering the possible abuse some nefarious users may try.

ApacheThunder:
The "Team Boxes" that gen 7 uses don't store pokemon the same way as gen 6. There's 6 team boxes instead of 1 battle box and unlike the battle box they aren't used as additional storage in the PC. The Pokémon are "registered" to each slot in the team slot while the original mon remains in the normal 30 or so PC boxes. You can move those mons around in the PC and they remain registered in the team box. (unless you move them to your party in which case they become unregistered from any team boxes they are associated with)

Unlike gen 6, I think we can do without the ability to read them. Moving mons there doesn't make then inaccessible from PkHeX as they remain in their original box slot. So at the moment I don't see the rush to get this implemented anytime soon. :P

kwsch:
https://github.com/kwsch/PKHeX/commit/251dc9c46a68aa9251ab00d0d5db60c68dfef02e

PR #633, opened 2016-12-13
Beat filthy cheaters

Eskuero:
On seventh gen there's no longer a battle box where the Pokemon data is moved, we now instead have six fake box teams that just contains a reference to where the Pokemon is located in the PC using Box and Slot index.
This data is found at offset 0x4CC4 with a lenght of 72 bytes. Every set of two bytes conforms a link to one Pokemon, being the first one the slot index of the box whose index is stored in the second byte. Empty slots are always chopped with two terminators (FF FF).

Example here:
01 00 00 00 06 00 02 01 1C 05 1D 1F FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF

Team 1 is using Pokemon from Slot 2 (Index 01) of Box 1 (Index 0), from Slot 1 of Box 1, from Slot 7 of Box 1, from Slot 3 of Box 2, from Slot 29 of Box 6 and from Slot 30 of Box 32. Every other Team is empty.

We check against any version other than Sun/Moon and enter a loop to cover every byte in the lenght of the Team Box data. We must move one index forward on the array after getting slot info to get the corresponding box index. If any box index matches the box we're on we check the slot index and if it matches as well we show an alert and stop the insertion of the data.

This is mainly made with the intention of prevent as much people as posible to try to cheat on official tournaments. There's nothing to do about people with hexadecimal editor maneuvering, CFW with QR injections and old binaries of PkHeX, but is still better than nothing.

### DIFF
--- a/PKHeX/MainWindow/Main.cs
+++ b/PKHeX/MainWindow/Main.cs
@@ -3229,6 +3229,19 @@ namespace PKHeX
         {
             if (!verifiedPKM()) return;
             int slot = getSlot(sender);
+            if (SAV.Version == GameVersion.SN || SAV.Version == GameVersion.MN) {
+                for (int i = 0; i < 72; i++) {
+                    int lslot = SAV.getData(19652 + i, 1)[0];
+                    i++;
+                    int lbox = SAV.getData(19652 + i, 1)[0];
+                    if (lbox == CB_BoxSelect.SelectedIndex) {
+                        if (slot == lslot) {
+                            Util.Alert("Failed to overwrite slot because is used by a Battle Box. Must take it out of team manually in-game.");
+                            return;
+                        }
+                    }
+                }
+            }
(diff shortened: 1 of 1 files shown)


kwsch:
This only handles clickSet, not the many other methods of data insertion :P

Refer to #544

Eskuero:
Ok, I'll make the loop in a different function so is less messy to call it from different places. Checking the clone to all slots in box should be fairly easy, but the drag and drop one is feeling like hell :')

Eskuero:
I always mess up Pull Requests sorry, check the new one instead as I close this.

PR #637, opened 2016-12-13
Beat filthy cheaters

Eskuero:
On seventh gen there's no longer a battle box where the Pokemon data is moved, we now instead have six fake box teams that just contains a reference to where the Pokemon is located in the PC using Box and Slot index.
This data is found at offset 0x4CC4 with a lenght of 72 bytes. Every set of two bytes conforms a link to one Pokemon, being the first one the slot index of the box whose index is stored in the second byte. Empty slots are always chopped with two terminators (FF FF).

Example here:
01 00 00 00 06 00 02 01 1C 05 1D 1F FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF

Team 1 is using Pokemon from Slot 2 (Index 01) of Box 1 (Index 0), from Slot 1 of Box 1, from Slot 7 of Box 1, from Slot 3 of Box 2, from Slot 29 of Box 6 and from Slot 30 of Box 32. Every other Team is empty.

We check against any version other than Sun/Moon and enter a loop to cover every byte in the lenght of the Team Box data. We must move one index forward on the array after getting slot info to get the corresponding box index. If any box index matches the box we're on we check the slot index and if it matches as well we show an alert and stop the insertion of the data.

This is mainly made with the intention of prevent as much people as posible to try to cheat on official tournaments. There's nothing to do about people with hexadecimal editor maneuvering, CFW with QR injections and old binaries of PkHeX, but is still better than nothing.

### DIFF
--- a/PKHeX/MainWindow/Main.cs
+++ b/PKHeX/MainWindow/Main.cs
@@ -3229,7 +3229,8 @@ namespace PKHeX
         {
             if (!verifiedPKM()) return;
             int slot = getSlot(sender);
-            if ((SAV.Version == GameVersion.SN || SAV.Version == GameVersion.MN) && IsLinkedSlot(slot, CB_BoxSelect.SelectedIndex)) {
+            if (SAV.IsLinkedSlot(slot, CB_BoxSelect.SelectedIndex)) {
+                Util.Alert("Failed to set to slot " + (slot + 1) + " of box " + (CB_BoxSelect.SelectedIndex + 1) + " because target is linked with battle box.");
                 return;
             }
             if (slot == 30 && (CB_Species.SelectedIndex == 0 || CHK_IsEgg.Checked))
@@ -3282,7 +3283,8 @@ namespace PKHeX
         private void clickDelete(object sender, EventArgs e)
         {
             int slot = getSlot(sender);
-            if ((SAV.Version == GameVersion.SN || SAV.Version == GameVersion.MN) && IsLinkedSlot(slot, CB_BoxSelect.SelectedIndex)) {
+            if (SAV.IsLinkedSlot(slot, CB_BoxSelect.SelectedIndex)) {
+                Util.Alert("Failed to delete from slot " + (slot + 1) + " of box " + (CB_BoxSelect.SelectedIndex + 1) + " because target is linked with battle box.");
                 return;
             }
             if (slot == 30 && SAV.PartyCount == 1 && !HaX) { Util.Alert("Can't delete first slot."); return; }
@@ -3390,7 +3392,8 @@ namespace PKHeX
 
             for (int i = 0; i < 30; i++) // set to every slot in box
             {
-                if ((SAV.Version == GameVersion.SN || SAV.Version == GameVersion.MN) && IsLinkedSlot(i, CB_BoxSelect.SelectedIndex)) {
+                if (SAV.IsLinkedSlot(i, CB_BoxSelect.SelectedIndex)) {
+                    Util.Alert("Failed to set to slot " + (i + 1) + " of box " + (CB_BoxSelect.SelectedIndex + 1) + " because target is linked with battle box.");
                     continue;
                 }
                 SAV.setStoredSlot(pk, getPKXOffset(i));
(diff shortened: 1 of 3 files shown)


Eskuero:
This prevents overwritting a slot linked with battle box:
- On click
- On cloning to all slots
- On drag and drop
- On delete
- On Box and full PC importing

I don't think I missed any possibility, but you tell me.

PKHeX/MainWindow/Main.cs
```diff
@@ -3279,6 +3282,9 @@ private void clickSet(object sender, EventArgs e)
         private void clickDelete(object sender, EventArgs e)
         {
             int slot = getSlot(sender);
+            if ((SAV.Version == GameVersion.SN || SAV.Version == GameVersion.MN) && IsLinkedSlot(slot, CB_BoxSelect.SelectedIndex)) {
```
evandixon: I think it would increase maintainability to move `(SAV.Version == GameVersion.SN || SAV.Version == GameVersion.MN)` to IsLinkedSlot, having the function return false if the version is not sun or moon.  The same applies to the 3 other uses in the same file.

PKHeX/MainWindow/Main.cs
```diff
@@ -3279,6 +3282,9 @@ private void clickSet(object sender, EventArgs e)
...
+                return;
```
evandixon: Maybe add a warning saying why the operation did nothing, something like this:
```
Util.Alert("Cannot delete; Pokemon inside battle box");
```

PKHeX/MainWindow/Main.cs
```diff
@@ -4304,7 +4315,21 @@ public static void Reset()
                 return slotSource == form || slotDestination == form; // form already updated?
             }
         }
+        public static bool IsLinkedSlot(int slot, int box) {
```
evandixon: I recommend moving this to `SaveFile` and `SAV7`.
For `SaveFile`:
```
public virtual bool IsLinkedSlot(int slot, int box) {
    return false;
}
```
And use the same logic for the override in `SAV7`.

PKHeX/MainWindow/Main.cs
```diff
@@ -4304,7 +4315,21 @@ public static void Reset()
...
+            for (int i = 0; i < 72; i++) {
+                int lslot = SAV.getData(19652 + i, 1)[0];
+                i++;
+                int lbox = SAV.getData(19652 + i, 1)[0];
+                if (lbox == box) {
+                    if (slot == lslot) {
+                        Util.Alert("Failed to modify slot "+ ++slot +" of box "+ ++box +" because is linked to a team. Take the Pokemon out of it in-game before trying again.");
```
evandixon: The function's name implies it simply determines whether or not a slot is linked.  As such, I don't think Util.Alert should be used here.

Eskuero:
I think I covered all of your suggestions but feel free to point some more stuff.
+1 Looks good to me, but someone else must approve.

ricksee:
@Eskuero What does this do exactly? How can players cheat on official tournaments with this?

Randactyl:
The team you choose for a tournament is locked - you can't change which Pokemon are registered to the team or the items they are holding. Without this change, PKHeX could be used to change Pokemon and/or items on a locked team.

PKHeX/Saves/SAV7.cs
```diff
@@ -1064,5 +1064,19 @@ public override bool RequiresMemeCrypto
                 return true;
             }
         }
+        public override bool IsLinkedSlot(int slot, int box) {
+            if (Version != GameVersion.SN && Version != GameVersion.MN) { return false; }
```
evandixon: Version check not needed because of inheritance.

Eskuero: Done

Eskuero:
Since this is a mess of commits should I close PR and just upload a clean commit with all the changes?

evandixon:
It's fine.  Pull requests should only ever be closed if you no longer want to merge the changes or if small fixes can't resolve concerns.

I approve the changes, but Kaphotics should be the one to merge the PR.

kwsch:
- Missing handling for BoxViewer, 
- Missing handling for setting Battle Videos, 
- Missing handling for clearing/sorting box(es)
- Intermixing WinForms code (Util.Alert) with save model (SaveFile.cs)

I'll end up rolling my own implementation that indicates if the slot is locked, as well as considering the cases changed here and concerns listed in my comment.

SydneySedai:
This change is locking my actual party slots, not sure if that's intentional.
Tried to delete an egg so I'm rather certain it's not linked in a battle team or anything.

Issue #618, opened 2016-12-09
Fixed the System.Deployment issue in Mono on macOS Sierra 10.12


jimga150:
I dont really know where to put this, but i want other people to know--if you comment out this line:

AddFontMemResourceEx(fontPtr, (uint)Resources.pgldings_normalregular.Length, IntPtr.Zero, ref dummy);

...in PKHeX/PKM/PKX.cs (line 601), and...

return System.Deployment.Application.ApplicationDeployment.IsNetworkDeployed;

in PKHeX/Util/DeployUtil.cs (then add a "return false;" under it), the project builds with only a few unused variable errors, and runs great. 

Heres what the edits look like in Xamarin:
![image](https://cloud.githubusercontent.com/assets/17803304/21062305/368b1550-be1e-11e6-9a6e-c101ce03ea5a.png)

![image](https://cloud.githubusercontent.com/assets/17803304/21062384/a0cc783c-be1e-11e6-9cfd-c384dbab6e4f.png)

Tsunamical:
https://github.com/julroy67/PKHeX/commit/95501698175f87dd719af235a30aba5181fead0f
https://github.com/julroy67/PKHeX/commit/fcc77fcf54f79950656be8d86a03bfe37c65a3f7

Better fixes.

evandixon:
You're not supposed to build with the ClickOnce-Debug or ClickOnce-Release build configurations.  Only Debug or Release.

PR #649, opened 2016-12-17
Build Configuration Clarification

evandixon:
Renamed build configurations to clarify usage (hopefully avoiding misconceptions like in #618).  Also added the README to the solution (so it can be easily modified from within VS) and modified it to briefly describe when to use each build configuration.

Commits 95501698175f87dd719af235a30aba5181fead0f and fcc77fcf54f79950656be8d86a03bfe37c65a3f7 from julroy67's fork can be adapted to apply only when Mono-Debug or Mono-Release is the active configuration, although depending on the exact implementation, a MONO compiler constant may need to be defined.


evandixon:
My team city server has been updated so that mono builds use the Mono-Release configuration.  This means that commits in other branches will report build failure until this PR is merged or until I change it back.

PR #689, opened 2017-01-01
HA Legality Check for Gen 7 Island Scan + Disable Unused Controls in Memories

ReignOfComputer:
Sun/Moon Island Scan Pokemon cannot have Hidden Abilities.


kwsch:
I had left the Memory editing interface unlocked so that people could modify the values if they really really wanted. I'm fine disabling it :)

ReignOfComputer:
You left it unlocked, but when people made any changes it wouldn't be persistent :P

PR #788, opened 2017-01-31
Update Update Path

ReignOfComputer:
Technically, should be changed to /pkhex/ because it's case sensitive (/PKHeX/ sends user to a 404), but I think it's best to just bring the user to the download page directly.


kwsch:
Probably easier to just add another redirect for the site so that existing copies are directed correctly.

evandixon:
I configured nginx to make the /pkhex be case-insensitive.  I think we should keep /pkhex around in case we ever change systems again or decide to change where it should redirect to.

PR #797, opened 2017-02-01
Catch MissingMethodException and Direct Users to .NET Framework 4.6 Install

ReignOfComputer:
Theoretically, this should work, though I don't have the means to test it now.

IMHO, perhaps the target framework should be 4.5 instead to extend support to Windows 8.0 users who have never updated their PC.


PKHeX.WinForms/Program.cs
```diff
@@ -93,6 +94,12 @@ private static void CurrentDomain_UnhandledException(object sender, UnhandledExc
                 // Todo: make this translatable
                 ErrorWindow.ShowErrorDialog("An unhandled exception has occurred.\nPKHeX must now close.", ex, false);
             }
+            catch (MissingMethodException)
+            {
+                // Todo: make this translatable
+                MessageBox.Show(".NET Framework 4.6 needs to be installed for this version of PKHeX to run.", "PKHeX Error", MessageBoxButtons.OK, MessageBoxIcon.Stop);
+                Process.Start(@"https://www.microsoft.com/en-us/download/details.aspx?id=48130");
```
evandixon: Have you tested the `Process.Start` in an environment that doesn't have .Net 4.6?  Quite frankly, I'm amazed that the code runs enough to get to the generic backup error message, but I wouldn't exactly trust any bit of code to work properly.

ReignOfComputer: `Process.Start` has been in use since .NET 2 and *should* work. We use it for opening the update thread anyway.

I haven't been able to test any of these as I don't have access to a vanilla Windows 7 PC right now.

evandixon: I'll set up a VM tomorrow then.  It probably _will_ work, but if things break, I don't want anyone saying "b-but you said...".

ReignOfComputer: Spun a VM. You're right, it doesn't work.

.NET Framework 3.5 (Vanilla Windows 7)
pkhex-win7sp1_3 5

.NET Framework 4.0
pkhex-win7sp1_4 0

.NET Framework 4.5
pkhex-win7sp1_4 5

Perhaps we should return to .NET 4.0? I'm able to use ILMerge with *.NET Framework v4 Client* on my Win 7 VM which would support all the OSes again.

evandixon: We should keep it on 4.6, since it comes with some new features. Windows XP support is no longer a priority, making it a bit easier.

The error for 3.5 is descriptive enough that we shouldn't have to worry about it.  The fact that the window shows in 4.5 means it's possible to display a nicer error, probably the same one from this PR, maybe even a varient of my ErrorWindow to allow copy/pasting the URL into the browser.  The real question is about 4.0.  Given that it normally throws a MissingMethodException, I'm guessing that it fails right on Process.Start?  If it shows the MessageBox, then that's probably plenty.  If not, we'll have to get creative, or make it painfully obvious in the forums that the .Net Framework 4.6 is needed.

ReignOfComputer:
Doesn't show the MessageBox. Went out for lunch, will do more experimenting when I'm back.

PKHeX.WinForms/Program.cs
ReignOfComputer: Doesn't show the MessageBox. Went out for lunch, will do more experimenting when I'm back.

evandixon: Time zones are weird.  I'm about to go to sleep, since it's 11:30PM.

Sort of a desperate attempt at giving user feedback, maybe if the application's underlying console was shown, with a message to get the correct framework version.  It would involve referencing kernal32, which should be independant from the framework.  Hopefully the Console class still works.

Simply call ConsoleManager.Show, then use the Console clas to write to it.  (C# conversion is easy, and I link to the source in the comments, which may still be valid):
https://github.com/evandixon/SkyEditor.Core/blob/master/SkyEditor.Core.Windows/Windows/Internal/ConsoleManager.vb

ReignOfComputer:
Okay, latest commit will now check for .NET Framework version on launch. This will only help users who are already on 4.5. Will additionally make a thread about this on the forum.

PKHeX.WinForms/Program.cs
ReignOfComputer: Think it's better to prevent those not on 4.6 from starting the application instead of try/catching or working unhandled exceptions since we don't know what other issues may arise from continued use.

PR #801, opened 2017-02-01
Tests: Use DateTime.UtcNow for test cases

Tsunamical:
DateTime.UtcNow is considerably faster than DateTime.Now, and since this
value isn't used for actual game purposes, we can take advantage of it.

### DIFF
--- a/Tests/PKHeX.Tests/PKM/PKMTests.cs
+++ b/Tests/PKHeX.Tests/PKM/PKMTests.cs
@@ -52,7 +52,7 @@ namespace PKHeX.Tests.PKM
             Assert.AreEqual(0, pk.MetYear, "Met_Year was not zeroed when MetDate is set to null");
 
             // Ensure setting to a date sets the components
-            var now = DateTime.Now;
+            var now = DateTime.UtcNow;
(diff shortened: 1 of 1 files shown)


evandixon:
The tests only use Date.Now because an arbitrary date is needed.  Changing to UtcNow has no benefits or drawbacks.

Tsunamical:
I understand that they use an arbitrary date - that's why I'm using it here and not for other places. Changing it has the benefit of increased performance with absolutely no drawbacks.

evandixon:
[You're right!](http://stackoverflow.com/questions/4075525/why-are-datetime-now-datetime-utcnow-so-slow-expensive) I wasn't aware that UtcNow was faster than just Now.

kwsch:
Consider applying the same changes to the other projects in the repo.

Tsunamical:
I'm not sure if we'd want that - hence why I held back on it. Say the user does it at 9 PM EST today, well in UTC it's the next day already so it'd automatically be set to that. It's up to you on that one.

kwsch:
Actually yeah that's fine then.

PR #802, opened 2017-02-01
Class consistency fixes

Tsunamical:
No functional differences, just better class typing.

### DIFF
--- a/PKHeX.WinForms/Util/NetUtil.cs
+++ b/PKHeX.WinForms/Util/NetUtil.cs
@@ -5,7 +5,7 @@ using System.Net;
 
 namespace PKHeX.WinForms
 {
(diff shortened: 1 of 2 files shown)


kwsch:
In general I prefer to keep uses of reflection obvious. Helps indicate that it's not just a general utility but it's using a more advanced language feature (and possibly slower than another possible implementation).

Tsunamical:
Okay, I'll definitely revert that then. That's a good idea.

NetUtil can be static and keeps it the same as the other utils for WinForms, and PGT should be sealed because of the virtual call, so I think those changes should still be acceptable.

Tsunamical:
Fixed, and better descriptions on each now.

PR #803, opened 2017-02-01
More code enhancements! [WIP]

Tsunamical:
WIP

### DIFF
--- a/PKHeX.WinForms/Subforms/PKM Editors/MemoryAmie.cs
+++ b/PKHeX.WinForms/Subforms/PKM Editors/MemoryAmie.cs
@@ -435,10 +435,10 @@ namespace PKHeX.WinForms
 
         private void update255_MTB(object sender, EventArgs e)
         {
-            MaskedTextBox mtb = sender as MaskedTextBox;
+            MaskedTextBox mtb = (MaskedTextBox) sender;
             try
             {
-                int val = Util.ToInt32(mtb?.Text);
+                int val = Util.ToInt32(mtb.Text);
                 if (val > 255) mtb.Text = "255";
             }
             catch { mtb.Text = "0"; }
--- a/PKHeX/PKM/PK7.cs
+++ b/PKHeX/PKM/PK7.cs
@@ -544,7 +544,7 @@ namespace PKHeX.Core
 
             if (GenNumber < 7) // must be transferred via bank, and must have memories
             {
-                TradeMemory(Bank: true);
+                TradeMemory(true);
                 // georegions cleared on 6->7, no need to set
             }
         }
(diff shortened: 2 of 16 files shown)


PKHeX.WinForms/Subforms/PKM Editors/MemoryAmie.cs
```diff
@@ -435,10 +435,10 @@ private void changeCountryText(object sender, EventArgs e)
 
         private void update255_MTB(object sender, EventArgs e)
         {
-            MaskedTextBox mtb = sender as MaskedTextBox;
+            MaskedTextBox mtb = (MaskedTextBox) sender;
```
evandixon: Casting using `as` is faster than type conversion.
https://www.codeproject.com/articles/8052/type-casting-impact-over-execution-performance-in

Tsunamical: That's an article written in 2004, in .NET Frameworks 2.0+, type conversion gained a significant performance increase. If the performance decrease is negligible and we can avoid using conditional access to prevent null dereferences, then I think in the long run that is the much better solution.

Since I don't see too many current benches, might as well do my own and post results, because I still see debate on this too.

evandixon: Type conversion is slower because it needs to consider user-defined conversion code if applicable, while type casting simply checks whether or not an object is a certain type.

Here's a newer article:
http://theburningmonk.com/2010/04/net-tips-using-as-vs-casting-in-csharp/

Tsunamical: Hm, I'll dig deeper into our usage, but regardless of the overall choice - there are some cases where we should be able to migrate one to the other safely.

PKHeX/PKM/PK7.cs
```diff
@@ -544,7 +544,7 @@ public void FixMemories()
 
             if (GenNumber < 7) // must be transferred via bank, and must have memories
             {
-                TradeMemory(Bank: true);
+                TradeMemory(true);
```
SciresM: Why remove the "Bank: "? It provides clarity with no meaningful downside....

Tsunamical: Oops, I actually didn't mean to push that, thanks for the catch. I meant to have that commit limited to casting changes, and I thought I took out everything I wasn't going to push, but I missed that.

Tsunamical:
I'm going to start from scratch on this one and look into uses where we can change type conversions to casting. I'll also look and see what some of our hottest functions are and if we can't speed some things up without compromising any feasibility.

Issue #807, opened 2017-02-02
Bug reports


sora10pls:
Currently, when exporting a Showdown set for Furfrou, it adds the Form name next to the Pokemon's name (example: `Furfrou-Star`). If this set is imported into Pokemon Showdown, the Pokemon is unknown as Showdown only accepts Natural Furfrou (listed as `Furfrou`); this means any and all Form names should be removed when exporting a Furfrou set.

What PKHeX exports:
![image](https://cloud.githubusercontent.com/assets/17801814/22561830/08a383a2-e948-11e6-81c9-45767c3bf7b5.png)

What happens when that set is imported into Showdown:
![image](https://cloud.githubusercontent.com/assets/17801814/22561853/2351328a-e948-11e6-9a35-12bb8bbff5a1.png)

Secondly, the new Giratina-Origin check is 50% accurate. If you create a Giratina in its Origin Forme that is not holding the Griseous Orb, the legality check is accurate. However, if you create a Giratina in its Altered Forme that **is** holding the Griseous Orb, it is deemed legal when it should be illegal (Altered Forme cannot have Griseous Orb without automatically becoming Origin Forme, and Origin Forme must have Griseous Orb).
![image](https://cloud.githubusercontent.com/assets/17801814/22561941/86359058-e948-11e6-9121-8b202396540b.png)

sora10pls:
I should also add the following for Vivillon-Poké Ball

PKHeX exports: `Vivillon-Poké Ball` 
Showdown exports: `Vivillion-Pokeball`

Like the one issue I brought up before regarding Ash-Greninja, Zygarde, and Minior, this is another case of Showdown listing Forms their own way

PR #810, opened 2017-02-03
Fix Showdown Import and Export Cases

ReignOfComputer:
Fixes https://projectpokemon.org/forums/forums/topic/40047-error-when-importing-pokemon-from-showdown/ with check to see if form is null first before trying to replace.

Fixes https://github.com/kwsch/PKHeX/issues/807 for Furfrou and Vivillion.


PKHeX/PKM/ShowdownSet.cs
```diff
@@ -146,6 +146,9 @@ public ShowdownSet(string input = null)
                 case 658: // Greninja
                     if (Ability == 210) Form += "-Ash"; // Battle Bond
```
egzonqj: Hey @ReignOfComputer since you are on this (just saw it while I was doing my own way), you should edit the Greninja case here and remove the "-", since it wont set the form properly from an imported showdown set. Maybe even the + should be removed?

ReignOfComputer: Yep that sounds right. Thanks.

PKHeX/PKM/ShowdownSet.cs
```diff
@@ -144,7 +144,10 @@ public ShowdownSet(string input = null)
             switch (Species)
             {
                 case 658: // Greninja
-                    if (Ability == 210) Form += "-Ash"; // Battle Bond
+                    if (Ability == 210) Form = "Ash"; // Battle Bond
+                    break;
+                case 666: // Vivillon
+                    if (Form == "Pokeball") Form = "Poké Ball";
```
evandixon: This check is there to translate the Pokemon showdown form name to the PKHeX form name.  I think it would be easier to just rename the form in text_Forms_en.txt, since the checks here and on line 284 could be removed.  Due to the nature of the change, however, @kwsch should approve this suggestion first.

kwsch: no, PKHeX mimics the games; Showdown does its own thing

Issue #773, opened 2017-01-30
Feature Request: Mass Legality Checker


PKMWM1:
One major addition that I think would be useful would be a mass legality checker. Similar to how Box Data Report compares various Pokemon Data against each other and other Pokemon, would it be possible to add the legality checker to one pf those columns, that we people don't have to check each individual Pokemon one-by-one and can know exactly which ones need to be looked at?

kwsch:
https://github.com/kwsch/PKHeX/issues/739

Issue #739, opened 2017-01-25
[Question] Is there a box-wide legality check? 


greenDarkness:
Sometimes when I open PKHeX it marks all my Pokemon as illegal, even the legal ones. If I few the Pokemon, and then click the red caution icon at the top it then gives me a legal notification. If I view the Pokemon again then it goes back to illegal. Viewing, and then setting the Pokemon fixes this issue. 

Below is a PK7 of an afflicted Pokemon.
http://www.mediafire.com/file/u3z1ooaobx7dmvx/132_%E2%98%85_-_Ditto_-_EE89382648C6.pk7

Thanks for reading!

kwsch:
database

When saving a PKM from tabs, PKHeX fixes common user errors and illegal values. Loading something to tabs and re-checking the legality causes PKHeX to save the PKM, fixing the illegal property.

Refer to https://github.com/kwsch/PKHeX/issues/604 item 3.

greenDarkness:
Is there a way to do that without loading each Pokemon into a tab and then checking the legality? 408 transferred Pokemon are effected by this, and it would take quite some time to load and save each one up manually. 

Thank you for answering my question.

kwsch:
that's the purpose of the batch editor

=HT_Gender=2
.HT_Gender=0

greenDarkness:
It says that it didn't modify any files. And they are all still marked as illegal. 
Alert
Modified 0/585 files.

kwsch:
![image](https://cloud.githubusercontent.com/assets/6393368/22282481/94ab7138-e291-11e6-9a60-0e86b14e9d07.png)

greenDarkness:
https://snag.gy/JK75Uv.jpg

kwsch:
Try setting a breakpoint [here](https://github.com/kwsch/PKHeX/blob/2cff51be61528bd62dec8b126507f5dce5e615f1/PKHeX/Legality/Analysis.cs#L49) and inspect what the Parse list contains in terms of legality analysis.

Loading a PKM should trigger it the first time, which should give a clue what is the actual error that is being automatically fixed => replicate in batch editor.

greenDarkness:
I'm not quite sure how to go about inserting a breakpoint, as currently I'm code illiterate. Would it be possible to dumb it down for me? Thanks so much for taking time to look at this!
[...]

Issue #299, opened 2016-10-02
Possible Option: Bulk Legality Check


ricksee:
Is it possible to do a mass legality check for all Pokemon in the boxes? If don't can that be added? Like it doesn't have to tell which problems all the Pokemon that didn't pass have. Like for Pokemon that are legal get a check on the left on their sprite. Those that don't get the red sign. It could also display how many passed and how many failed.

kwsch:
![img](http://i.imgur.com/loSbfCm.png)

PR #814, opened 2017-02-03
Allow to see at a glance (in the box) the mons that are illegal.

JortonMV:
When a pokémon is not legal changes the background of its slot.

Its only an idea, I thought that people will like to see at a glance if they have illegal pokemon, without having to see them one by one.

ss

### DIFF
--- a/PKHeX.WinForms/MainWindow/Main.cs
+++ b/PKHeX.WinForms/MainWindow/Main.cs
@@ -3402,6 +3402,7 @@ namespace PKHeX.WinForms
             updateBoxViewers();
 
             RedoStack.Clear(); Menu_Redo.Enabled = false;
+			checkIfLegalPKMSlot(pk, SlotPictureBoxes[slot]);
         }
         private void clickDelete(object sender, EventArgs e)
         {
@@ -3838,7 +3839,18 @@ namespace PKHeX.WinForms
             pb.Image = sprite;
             pb.BackColor = Color.Transparent;
             pb.Visible = true;
+			checkIfLegalPKMSlot(p, pb);
         }
+		private void checkIfLegalPKMSlot(PKM pk, PictureBox pb)
+		{
+			LegalityAnalysis la = new LegalityAnalysis(pk);
+			if (pk.GenNumber >= 6 && pk.Species != 0 && !la.Valid)
+			{
+				pb.BackColor = Color.FromArgb(150, Color.Red);
+			}
+			else
+				pb.BackColor = Color.Transparent;
+		}
         private void getSlotColor(int slot, Image color)
         {
             foreach (PictureBox t in SlotPictureBoxes)


evandixon:
I think this is a great idea, although the a colored background can be hard to see in some situations.
image

PKHeX.WinForms/MainWindow/Main.cs
```diff
@@ -3838,7 +3839,18 @@ private void getSlotFiller(int offset, PictureBox pb)
             pb.Image = sprite;
             pb.BackColor = Color.Transparent;
             pb.Visible = true;
-        }
+			checkIfLegalPKMSlot(p, pb);
```
evandixon: Nit-pick: The new lines of code use tabs instead of spaces, making them look more indented than needed.

PKHeX.WinForms/MainWindow/Main.cs
```diff
@@ -3838,7 +3839,18 @@ private void getSlotFiller(int offset, PictureBox pb)
...
+        }
+		private void checkIfLegalPKMSlot(PKM pk, PictureBox pb)
+		{
+			LegalityAnalysis la = new LegalityAnalysis(pk);
+			if (pk.GenNumber >= 6 && pk.Species != 0 && !la.Valid)
+			{
```
evandixon: Inconsistent use of curly braces.  The else section doesn't use them.  I personally would go with adding them to the else, but @kwsch (from what I've seen) would remove the curly braces here.

PKHeX.WinForms/MainWindow/Main.cs
```diff
@@ -3838,7 +3839,18 @@ private void getSlotFiller(int offset, PictureBox pb)
             pb.Image = sprite;
             pb.BackColor = Color.Transparent;
             pb.Visible = true;
-        }
+			checkIfLegalPKMSlot(p, pb);
+        }
+		private void checkIfLegalPKMSlot(PKM pk, PictureBox pb)
+		{
+			LegalityAnalysis la = new LegalityAnalysis(pk);
+			if (pk.GenNumber >= 6 && pk.Species != 0 && !la.Valid)
```
evandixon: For performance in other generations, it would be better to check `pk.GenNumber` and `pk.Species` before creating the legality analysis.

egzonqj:
I think it would be better to show a little red triangle on the slot instead of changing its background color

kwsch:
I'd prefer to not add more visual clutter to the box slots; we already have:
* Species
* Shiny
* Held Item
* Battle Box Slot / Lock

This would slow down flipping through boxes by a small amount, as instead of just loading and displaying data you're now checking the legality of every slot.

The Database tool already has the option of searching and displaying illegal PKM in the box, and you can directly delete them (updates the box).

image

https://github.com/kwsch/PKHeX/issues/299
https://github.com/kwsch/PKHeX/issues/739
https://github.com/kwsch/PKHeX/issues/773

PR #818, opened 2017-02-05
Check Illegal slot (redo of previous PR from Jorton)

egzonqj:
This is based on @JortonMV 's PR/idea, just did it to match what @kwsch said about it being a Set to SAV option, would like a hard review here, since I don't really know if how I did was the best path to do it. 

Also changed the background-color modification to a little image on the right side.

image


egzonqj:
Just noticed that some of the commented text that are autogenerated by VS are now in spanish, truly sorry for that.

PKHeX/Saves/SaveFile.cs
```diff
@@ -8,6 +8,7 @@ public abstract class SaveFile
     {
         public static bool SetUpdateDex = true;
         public static bool SetUpdatePKM = true;
+        public static bool ChkIllegal = true;
```
evandixon: This variable is set but never used.  It (and the things setting it) can be safely removed.

evandixon:
Looks good!
image

Nit pick: It can be hard to see in some cases, but I don't recommend changing it.
image

PR #830, opened 2017-02-06
Remove excess line

Mrtops:
Get rid of extra blank line


kwsch:
It's a split between the two types of records stored, 0-99 are 32bit and 100-199 are 16bit.

https://github.com/kwsch/PKHeX/blob/c401781526c62191f3e242f60b5c155946771be2/PKHeX/Saves/SAV7.cs#L593-L626

Mrtops:
Oh ok

PR #836, opened 2017-02-08
Fix Coin Getter in Gen 1

ReignOfComputer:
Fixes: https://projectpokemon.org/forums/forums/topic/40119-another-trainer-info-error-this-time-for-gen-1


Tsunamical:
For future reference, if you mess up the commit, please just git reset HEAD~x (x being amount of commits to roll back), commit and git push -f. That way you're not opening and closing a new PR every time.

ReignOfComputer:
Thanks @Tsunamical. Still figuring out Git :)

PKHeX/Saves/SAV1.cs
```diff
@@ -328,7 +328,7 @@ public uint Coin
         {
             get
             {
-                return uint.Parse(BigEndian.ToUInt16(Data, Japanese ? 0x2846 : 0x2850).ToString("X4"));
+                return Convert.ToUInt16(BigEndian.ToUInt16(Data, Japanese ? 0x2846 : 0x2850).ToString("X4"), 16);
```
evandixon: I'm probably missing something, but why not just return `BigEndian.ToUInt16`?, instead of turning it to a string then converting it back?

PR #838, opened 2017-02-09
Compress Pokemon sprites to save space

Mrtops:


kwsch:
https://github.com/kwsch/PKHeX/pull/517

Please provide:
* Description of changes (tools used),
* Confirmation that sprites appear properly for mono and windows,
* Confirmation that filesize (dll AND compiled zip) was actually decreased (before/after).

No point in changing something that isn't broken; it isn't y2k where disk storage is a premium.

Mrtops:
Ok will provide info then get back

Mrtops:
compresspng.com was used to compress images
I checked and sprites do appear properly
File size does shrink by approx. 100Kb

Mrtops:
That was incorrect file size of the dll shrinks by about 20 kb

Tsunamical:
You'd probably get even better compression with ECT or zopfliPNG, but it's really not worth when it's more prone to cause problems like last image compression did.

Mrtops:
What happened last time?

evandixon:
Latest build of [`master`](http://teamcity.projectpokemon.org/viewLog.html?buildId=773&buildTypeId=PKHeX_BuildWindows&tab=artifacts)
Latest build of [`838/merge`](http://teamcity.projectpokemon.org/viewLog.html?buildId=797&tab=buildResultsDiv&buildTypeId=PKHeX_BuildWindows) (aka this one)

The size of the resulting zip file actually increases by 0.05 MB.  Both builds took place on the same build agent.

PR #845, opened 2017-02-10
Compress all images to save space

Mrtops:
I have compressed al images to save space I have checked and the program works fine on windows and mono. The images do display properly as well. I used PNGGauntlet to compress them.
The unzipped pkhex.core.dll size of kwsch master is 5827Kb
The zipped pkhex.core.dll size of kwsch master is 2630Kb
The unzipped pkhex.core.dll size of mrtops master is 5713Kb
The zipped pkhex.core.dll size of mrtops master is 2585Kb


ReignOfComputer:
Is there a point to saving 114kb? That's really not a lot.

SciresM:
We've had bad experiences with image compression in the past. Best not to mess with it.

PR #852, opened 2017-02-13
Add Framework check for non-windows OS

spof95:
GetFrameworkVersion() throws an System.NullReferenceException since
there is no Registry on non-windows systems.

Fixed by checking the OS under which the program runs. If it's not
Windows, check the CLR Version instead of querying the (non-existant)
registry.

### DIFF
--- a/PKHeX.WinForms/Program.cs
+++ b/PKHeX.WinForms/Program.cs
@@ -29,16 +29,33 @@ namespace PKHeX.WinForms
 
             try
             {
-                if (GetFrameworkVersion() >= 393295)
-                {
-                    StartPKHeX();
-                }
-                else
-                {
-                    // Todo: make this translatable
-                    MessageBox.Show(".NET Framework 4.6 needs to be installed for this version of PKHeX to run.", "PKHeX Error", MessageBoxButtons.OK, MessageBoxIcon.Stop);
-                    Process.Start(@"https://www.microsoft.com/download/details.aspx?id=48130");
-                }
+				if (IsOnWindows())
+				{
+					if (GetFrameworkVersion() >= 393295)
+					{
+						StartPKHeX();
+					}
+					else
+					{
+						// Todo: make this translatable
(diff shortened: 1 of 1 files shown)


kwsch:
https://msdn.microsoft.com/en-us/library/3a8hyw88(v=vs.110).aspx

Looks good, tabs vs spaces is nbd.

spof95:
To comment on IsOnWindows, as I saw your link, to retain compatibility with older releases of .NET, I have to use it that way since there are older versions of .NET which don't have that struct and just assigned 128 to UNIX. Therefore we need to compare to int litterals instead of the enum.

PKHeX.WinForms/Program.cs
```diff
@@ -29,16 +29,33 @@ private static void Main()
...
+                        Process.Start(@"https://www.microsoft.com/download/details.aspx?id=48130");
+                    }
                 }
                 else
                 {
-                    // Todo: make this translatable
-                    MessageBox.Show(".NET Framework 4.6 needs to be installed for this version of PKHeX to run.", "PKHeX Error", MessageBoxButtons.OK, MessageBoxIcon.Stop);
-                    Process.Start(@"https://www.microsoft.com/download/details.aspx?id=48130");
+                    //CLR Version 4.0.30319.42000 is equivalent to .NET Framework version 4.6
+                    if ((Environment.Version.CompareTo(Version.Parse("4.0.30319.42000"))) >= 0)
```
evandixon: I'm still testing in VMs, but this works in Windows 10, even after locally switching the project to .Net 4.6.1 and .Net 4.6.2.  You might not even need the OS check.

spof95: Microsoft discourage the use of Environment Version to check for the .NET Framework Version as the CLR version doesn't necessary have to change. It just happened to be a clear version number for 4.6: https://msdn.microsoft.com/en-us/library/hh925568(v=vs.110).aspx#clr_b 

As I am only able to check on Arch (kernel 4.9.5) and Win 10, I chose the safer route.

PR #856, opened 2017-02-13
Make wc* and pk* display properly on readme.md

Mrtops:


kwsch:
For markdown, the correct escape character is "\"; your commit yields a space between the asterisk and the close parenthesis.
I'll update the readme with the necessary terminator characters, thanks for reporting :+1:

Mrtops:
Ok I could no figure out how to get to work so I just did my best and thanks for the info

PR #858, opened 2017-02-14
Menu sprite distortion fix

### LINKED ISSUE #857: Disorted pokemon images in latest commits
Hi,

In gen 6 and 7 games which I have tested, some pokemon have their image disorted,more like zoomed in.Below are the images showing that:

In the Gen 6 AS main.sav file,this happens to some .pk5(shiny metagross-event from white) and .pk6 files(the rest 3 shinies). The other 4 above them are the same pokemon without being shiny(I have generated them just to show you the difference),the image is normal on them. When I change them again to shiny they gain the disorted image again:

In Gen 7 now the imported sableye, which is from my AS save and transfered to Moon via Pokemon Bank, the image is disorted without being shiny and changing the shiny mark doesn't affect it, it retains the same image:

The .pk7 sableye that I have in my box looks normal without it's shiny form unlike the .pk6 imported from my AS.Changing it though from shiny to non shiny the image gets disorted again:

This bug only occurs in the latest commits not the stable version (2017.02.07)
Thank you

sora10pls:
Revert OR/AS Steelix to X/Y; fixes sprite distortions if an image were overlaid onto the sprite (i.e. shiny star, held item, or legality flag when set to SAV).

Resolves #857


sora10pls:
Picture for reference:

image

PR #863, opened 2017-02-16
Past generations legallity tweaks

javierhimura:
4 months ago i started to create files to help making past generation legal checks, the levelup, eggmoves and tmhm tables.

Back then it can yet be used because PKHeX was not ready.
Now with the last refactors made to the program I have resumed the extraction and i have also the encounter tables for gen 3 and 4. When i complete the extraction of encounters for gen 5 and evolution tables for gen 3 to 5 i will make a pull request to include it in PKHeX

For now i made some tweaks to the legalliy checks that would make more easy to start those past gen legal chechks.

The most important one is to generate different evolution chain arrays depending on the generation, to allow for a given pokemon to know the evolution chain across all the generation that it has inhabited, chaind that can be different not only because of future gen evolutions, but also to take into account what if the max level in what the pokemon inhabited every generation, that would make some of the evolutions phases impossible in some generations, if a level 100 XY pokemon was originated in gen 3 and transfered to gen 4 with level 40... it coldnt learn any move above level 40 in gens 3 and 4

The second important change is to use met level to filter encounter slots even if the original met encounter is lost. Okay, in those cases some information is lost, but that not means every encounter slot is valid, the pokemon was captured with level equals or bellow the met level, that exclude every encounter with min level greather that met level

### DIFF
--- a/PKHeX/Legality/Structures/EvolutionTree.cs
+++ b/PKHeX/Legality/Structures/EvolutionTree.cs
@@ -416,22 +416,25 @@ namespace PKHeX.Core
                         continue;
 
                     oneValid = true;
-                    if (evo.Level == 0 && !evo.RequiresLevelUp) //Evolutions like elemental stones
+                    if (evo.Level == 0 && !evo.RequiresLevelUp) //Evolutions like elemental stones, trade, etc
+                    {
                         dl.Last().MinLevel = 1;
+                    }
                     else if(evo.Level ==0) //Evolutions like frienship, picho -> pikachu, eeve -> umbreon, etc
                     {
                         dl.Last().MinLevel = 2;
-                        if (dl.Count() > 1 && dl.First().MinLevel == 1)
-                            dl.First().MinLevel = 2;
+                        if (dl.Count() > 1 && !dl.First().RequiresLvlUp)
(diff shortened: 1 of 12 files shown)


PKHeX/Legality/Structures/EvolutionTree.cs
```diff
@@ -416,6 +416,25 @@ public IEnumerable<DexLevel> getExplicitLineage(PKM pkm, int lvl, bool skipCheck
...
 
                     oneValid = true;
+                    if (evo.Level == 0 && !evo.RequiresLevelUp) //Evolutions like elemental stones, trade, etc
+                    {
+                        dl.Last().MinLevel = 1;
+                    }
+                    else if(evo.Level ==0) //Evolutions like frienship, picho -> pikachu, eeve -> umbreon, etc
+                    {
+                        dl.Last().MinLevel = 2;
+                        if (dl.Count() > 1 && !dl.First().RequiresLvlUp)
```
evandixon: Nit pick: It's better to use `List<T>.Count` instead of the Linq `Count()` extension.  The extension is optimized so that the difference is negligible in this case (since `dl` is a List), so I'd only recommend changing it if you have to change something else.

kwsch:
Thanks :)

I'll be making changes for style tweaks / extra analysis in commits over the next week.

PR #877, opened 2017-02-22
Add .gci in the supported file type dropdown

LegoFigure11:


kwsch:
Should instead throw it with the Save Files option (`"|Save Files|*.sav;*.dat;"`); no need to separate the two. Main Saves are worthy of their own separate group as they are current generation and extensionless.

PR #905, opened 2017-02-28
add FormList to Gen7 Pokedex Editor 

pokecal:
Fixed: Wishiwashi schoolForm will be checked when use allModifying button.
Show mons by formName, I think it will be easier to observe the dex data.


kwsch:
very cool!

PR #918, opened 2017-03-03
Update PKHeX.sln

Mrtops:
Change Visual Studio version to 15 and add description


kwsch:
Unnecessary ;)

Deletes the new configuration options that were recently added, and the .sln is a very minor IDE thing.

PR #921, opened 2017-03-04
add batch PID randomize(keep shiny)

pokecal:
I think the constant value requires more meaningful name. I just concatted the existing 2 names.


pokecal:
I found the require operation and "=IsShiny=True" is available, just now.

PR #930, opened 2017-03-09
Check for Null Event Constant Difference before Clipboard Export

ReignOfComputer:
Prevents ArgumentNullException.


ReignOfComputer:
Fixes: 
image

PR #947, opened 2017-03-19
Gen 3 and 4 Encounter Slots, Gen 3 to 5 Evolves

javierhimura:
I extracted gen 3 and 4 encounters and 3 to 5 evolves from the gba and nds games.
The evolves and gen 4 encounters data is in the raw format from the original games, i ony added location if before the array of data from an encounter area.

Encounters data from gen 3 is almost raw data, but because the data in the gba rom is not in the same are,  it use many pointers, i joined all the needed data wihtout the pointers. I added a byte with 0 or 1 to determine if an encounter type exits, in gba a 24 bit pointer is used with 0 in case the data does not exits.


javierhimura:
I fortgot to mention it.

The encounters data contains all the normal encounters.
It is missing data from G4 Safari Zone because is not stored in the normal encounters table.
Is also missing data from Honey Trees and Headbutt in G4 and. I am in the process of extract headbutt encounters from the NDS rom.

javierhimura:
Well , forgot what i said about honey trees and heabutt, it is included.
The only thing missing is the safari zone from gen 4. I cant find any information about how the safari zone data is stored in the rom to extract it, i think safari encounters are too many encounters too manually add as alt slots

kwsch:
For PIDIV-encounter slot checking the encounter slots will eventually have to be marked with their slot type (grass/water) as well as their RNG % (0-19, 20-39, etc).

For future optimization we can precompute some of these alterations so that a lot of the set-up logic is already done. Getting the raw info is the biggest hurdle though, so thanks :D

PR #953, opened 2017-03-20
Complete gen 3 and 4 tutor moves

javierhimura:
Get tutor moves for gen 3 and 4 pokemon, including special tutors
Filter deoxys gen 3 level moves by form
Remove HMs in gen 3 and 4 when transfered to another generation, even if gen 3 HMs are not added some pokemon can learn HMs by leveling up
Defog and Whirlpoll are not removed because a pokemon can be transfered to gen 5 learning only one, should be analyzed that a pokemon in gen 5 with both moves is illegal

### DIFF
--- a/PKHeX/Legality/Core.cs
+++ b/PKHeX/Legality/Core.cs
@@ -2079,9 +2079,22 @@ namespace PKHeX.Core
                         if (Machine)
                         {
                             var pi_hgss = PersonalTable.HGSS[index];
-                            r.AddRange(TMHM_HGSS.Where((t, m) => pi_hgss.TMHM[m]));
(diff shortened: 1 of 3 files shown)


javierhimura:
Also i fixed gen 4 TMHM tables, personal tables for gen 4 does not caints at the same time data for Defog and Whirlpoll compatibility. TM and HM are separated now like in gen 3, with the exception that Defog and Whirlpoll should be added always

kwsch:
The transfer 4->5 forbids having both moves at the same time; I'm sure there are some species that can learn both so a special check would have to be added to ensure that both can't be present at the same time (unless possible to learn in gens 5/6/7). More of a fine-tuning check to do once everything else is implemented.

PR #963, opened 2017-03-21
Misc table fixes

wwwwwwzx:
- fix frlg/dppt roaming locations table after testing
- fix previous typo


wwwwwwzx:
Also fixed a bug when I was trying to only keep the pokedex of suicune xD

javierhimura:
Unown A is commented because it has form 0. Form 0 is the only form loaded from encounters raw file because gen 4 encounters file does not contains alternate forms info.

But unlike gen 3 all the unown forms are found in the same area, that make it possible to just load form 0 from the raw files and only add manually non 0 forms.

wwwwwwzx:
I see. Thank you for pointing out

PR #966, opened 2017-03-22
Add B2W2 hidden grotto table

wwwwwwzx:
todolist:
- legality checking of the ability of pkm from hidden grotto (must be HA)
- check the levels of pkm obtained from Funfest Missions


kwsch:
Feel free to add another slot type to the enum, preferably indicating that HA is required for later inclusion as an ability check. (kinda like the Friend Safari hidden ability check)

PR #968, opened 2017-03-22
fix gen4dex forms

pokecal:
mismatch of formFlag, and languageFlag also fixed.
full edit of DexUpgrade added.
DP,with SetToSavOption, some mons otherLangDexFlag will be edit.


pokecal:
I confirmed on Pearl, HeartGold.
But I don't have Platinum, so the DexUpgrade not works on Pt.

PR #971, opened 2017-03-23
Start Rolling Back Some Direct Casts

Tsunamical:
We can safely roll these back, there are likely more cases like this,
but most of these cases were changed in
de8cebf2a4a872f2eab688d73e85203ff0b47899, which can actually be
reverted.


kwsch:
direct cast is preferred since we know that it should be called correctly. No need to null check.
http://stackoverflow.com/questions/132445/direct-casting-vs-as-operator

PR #973, opened 2017-03-23
Impl vivillon form legality checking

wwwwwwzx:


PKHeX/PKHeX.Core.csproj
```diff
@@ -711,10 +712,8 @@
     <None Include="Resources\text\en\text_EncounterType_en.txt" />
     <None Include="Resources\text\de\text_EncounterType_de.txt" />
     <None Include="Resources\text\locale\regions3ds.txt" />
-    <None Include="Resources\text\locale\3dsregions.txt" />
```
evandixon: Is removing these files from the project intentional?  I don't see the files being deleted, so it looks accidental.

wwwwwwzx: these files were renamed but forgot to deleted in the past I think, when I was looking for country and region ID I notice that.

PR #977, opened 2017-03-23
Fix addExtraTableSlots for areas only in extra table

javierhimura:
The function addExtraTableSlots does not take into account areas that exits in the extra areas array but not in the game areas array. There is many of this kind of areas in gen 1, 2 and some in gen 3 and 4 (For example Floaroma Meadow with a honey tree but not any normal wild encounter.

With the change it will continue to add the extra slots to the slots in the game areas if the location exits in both arrays but it will also add to the output the areas that only exits in the second array.

Also i fix one of the locations of shellos alt forms


javierhimura:
I do not know why the first 4 commits are included in the pull request, those commit are already in the official master branch.

Also the commit of legality_en was made because for some reason the last version of the file have the end line chars with a invalid encoding and git was changing it every time i tried to download the file, even if i reverted the file it was inmediatly changed. I couldnt change branchs without commit the file first.

PR #980, opened 2017-03-24
Add gen5 Dream World table

wwwwwwzx:
- Add a lengthy dream world table with special moves and levels (wonder if there is already a binary database). I guess location can be a good flag from PDW.
- Fixed a minor typo at vivillontables.
- Working on adding PGL Gift from Dream World (finished)


wwwwwwzx:
it seems encounters of N's Pokemon is the last thing in gen5 (except events). Treat them as statics? slots? or gifts?

kwsch:
encounterstatic is preferred, not as a gift unless the ball can be only one ball

there's already an NSparkle flag for encounterstatic

PR #988, opened 2017-03-25
Add PokeWalker encounters

wwwwwwzx:


wwwwwwzx:
I wonder if it's better to use reference to mark the location instead of using the clone method..

javierhimura:
Is better to mark the location in the proper static object like any other encounters only possible in one location and leave clone method only for static locations with more that one met location (roaming only i think)

javierhimura:
Also Gift = true should be added to pokewalker encounters, to mark illegal pokewalker pokemons that are not in pokeball

wwwwwwzx:
I guess so. Thank you

PKHeX/Resources/text/zh/LegalityCheckStrings_zh.txt
```diff
@@ -145,7 +145,6 @@ V111 = 非帮手宝可梦拥有隐藏特性。
 V300 = 在非群战/朋友原野区获得隐藏特性。
 V112 = 隐藏特性不可获得。
 V217 = 隐藏洞穴捕获的宝可梦必定为隐藏特性。
-V218 = 在野外普通相遇不可能有隐藏特性。
```
evandixon: Why did you remove this line?

kwsch: Doesn't currently exist in LegalityCheckStrings.cs :)

PR #990, opened 2017-03-25
Crossgen Moves legal analysis and pokemon box encounter

javierhimura:
parseMoves change to check legal moves per generation allowing the legal analysis to specify if a legal move learned is from current generation or past generations with more detailed info.

Also if the pokemon have a gen 3 or 4 HM that should have been removed it gives a error indicating that, is better that remove HM from the valid moves set, this analysis will also detecte non-remove HM from static encounters or events (pichu with surf from gen 3)

New legal edge cases detected.
- Gen 4 pokemon transfer to 5 with Defog and Whirlpool
- Gen 2 pokemon with non-tradeback egg moves and generation 1 exclusive moves. To trade a pokemon to gen 1 games every move that does not exits in gen 1 should be removed, that means is impossible to combine gen1 moves and gen 2 egg moves above MaxMoveId1
- Gen 3 pokemon from special eggs, from now pokemon box only but it could include in the future gen 3 events eggs like wish chansey or gen 2 event eggs (if exits)
- With the last inclusion of special eggs it also detects combination of moves from special eggs and normal eggs, are not possible

Also in included static encounters for pokemon box and added in game egg gift and fatefull encounter (used by pokemon box) to possible egg met location in gen 3 and modified the function getEggMoves for gen 2 to make it like the function of get encounters


javierhimura:
The only cross gen check i can think that is not include is gen 1 initial moves with egg moves. Because there is no move reminder in gen 1 if a pokemon have egg moves it should not have moves from gen 1 learned before level 6.

PR #993, opened 2017-03-26
Fix gen4 encounters reading

wwwwwwzx:
- fix selection from encounter static tables marked with grouped GameVersion (DP was not recognized as D or P)
- fix hgss sound slots reading
- Add Spiritomb encounter


wwwwwwzx:
It seems Sounds Slots replace 2,3,4,5 four slots.

javierhimura:
Contains method should also included special gametypes.
There are encounters in gen 1 and 2 marked with game special and pokemon box in gen 3

javierhimura:
About sound encounters to replace correctly slots 2,3,4 and 5 this
internal static readonly int[] Slot4_Sound = {4, 5};
Should be chaned into this in tables4.cs
internal static readonly int[] Slot4_Sound = {2, 3, 4, 5};

Does not overlap with time replaces (2,3) Time replace is used in DPPt and Sound replace in HGSS

Source:
https://projectpokemon.org/forums/forums/topic/273-ppre-project-pokemons-rom-editor/?page=25#comment-153893

javierhimura:
About the change in the function g_replace is used by both hgss and dppt but the dppt store the species in 4 bytes instead of 2, maybe a species lenght param should be added?

wwwwwwzx:
It's not elegant, but it works.
I saw a `National Radio` at last, is that swarming?

wwwwwwzx:
My reference is [here](https://sites.google.com/site/pokemonslots/gen-iv)
it seems fishing and surfing swarming works differently

javierhimura:
The National Radio is something that i found in the narc encounters files definition from projectpokemon
http://web.archive.org/web/20140802044813/http://projectpokemon.org/rawdb/heartgold/enc.php
In the "format" link it ends with

168	1	superrod1minlevel
169	1	superrod1maxlevel
170	2	superrod1natid
172	1	superrod2minlevel
173	1	superrod2maxlevel
174	2	superrod2natid
176	1	superrod3minlevel
177	1	superrod3maxlevel
178	2	superrod3natid
180	1	superrod4minlevel
181	1	superrod4maxlevel
182	2	superrod4natid
184	1	superrod5minlevel
185	1	superrod5maxlevel
186	2	superrod5natid
188	2	radionatid1
190	2	radionatid2

I first thought it was another radio listened after obtaining the national dex but maybe it is information about swarms
http://bulbapedia.bulbagarden.net/wiki/Radio_in_the_Pok%C3%A9mon_world#Professor_Oak.27s_Pok.C3.A9mon_Talk
If that is true swarms are loaded twice, we should look into binary file to check if that data is swarm for hgss, if it is the swarms loaded in binary files can be discarted and load only the manual list (it will be a mess to look while loading the binary file if the swarm is fish, water or grass when that data is not present in the file)

javierhimura:
Your source have exactly the same info about sound replaces, it should be

internal static readonly int[] Slot4_Sound = {2, 3, 4, 5};

The only thing left is to determine if NationalRadio is another sound replace (i dont think so anymore) or just swarms

wwwwwwzx:
Here is the last several bytes of route 1 to route 46.
Sometimes it is the swarming species but sometimes has other meaning

javierhimura:
Route 1 have Poochyena as swarm encounter
http://bulbapedia.bulbagarden.net/wiki/Kanto_Route_1#Generation_IV
Route 3 have Baltoy as swarm encounter
http://bulbapedia.bulbagarden.net/wiki/Kanto_Route_3#Generation_IV

PR #996, opened 2017-03-26
Gen 3 and 4 Safari Zones and Bug Contest Encounters

javierhimura:
Included encounter tables for HGSS Safari Zone and Bug Contest. Gen 3 and DPPt Great Marsh are included in the raw encounters files but this HGSS encounters were not.

Also added verification when there is more than one encounter possible after transfering gen 3 and 4 pokemon to gen 4 and 5. Encounters that does not match the ball are only discarted when there is at least one valid encounter, if there is only one possible encounter it will be returned as valid encounter and later in verifyBall (it is not yet implement for pre-gen6 encounters) the ball will be marked as illegal for the encounter.

Also gen 3 and 4 safari and bug contest have a special slot type marked to make in the future the verifyBall  checks.


kwsch:
Will eventually need to note the encounter slot number for PIDIV legality, but it's a good start to get the encounter info first :)

Bug Catching Contest:
http://www.smogon.com/forums/threads/past-gen-rng-research.61090/page-13#post-2984739

Safari Zone: (link 404'd)
http://www.smogon.com/ingame/rng/dpphgss_rng_part2#safari

Better Safari Zone source (no idea where it is in the ROM to verify):
http://upcarchive.playker.info/0/upokecenter/content/pokemon-heartgold-and-soulsilver-safari-zone-pokemon.html

PR #997, opened 2017-03-26
Gen3 WasEgg Pokemon in gen 4 or later, Gen 4 transfer encounter

javierhimura:
pkm.WasEgg is never true for gen 3 pokemon because there is no egg data to identify. Like gen 2 assume every pokemon above level 5 in pokeball that can be hatched from an egg that are eggs

For this pokemon check always for an egg encounter first, if does not match egg encounter then check other encounters

When verify moves for this pokemon it here is a non egg possible encounter first check moves from that encounter and if does not match check egg moves

Also for gen 2 unhatched eggs ignore gen 1 encounters

### DIFF
--- a/PKHeX/Legality/Checks.cs
+++ b/PKHeX/Legality/Checks.cs
@@ -738,8 +738,19 @@ namespace PKHeX.Core
 
                 EncounterMatch = null; // Reset Encounter Object, test for remaining encounters
             }
-            
-            if (pkm.WasEgg)
+
+            if(pkm.Gen3 && !pkm.HasOriginalMetLocation)
+            {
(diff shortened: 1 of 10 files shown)


javierhimura:
I added a similar function to check gen 4 encounters from pokemon in gen 5 or later.
This is almost like normal encounter analysis with the addition of check for invalid transporter location, in case there ir no other error in the encounter location but transporter location is invalid it return transporter error

javierhimura:
Added check to explicitely detect unreleased events
- Non Japanese Faraway Island Mew
- Hall of Origin Arceus
- Diamond/Pearl Flower Paradise Shaymin

PR #1006, opened 2017-03-28
Two minor fixes

wwwwwwzx:
- fix wc7 to pkm for egg: event egg should be nicknamed, not sure how gen6's events work
- fix legality check for in-game trade: distinguish two locations with the same name
Thanks to RoC's repo, complete or corrected some datapoint


PKHeX/Legality/Core.cs
```diff
@@ -927,7 +927,7 @@ internal static EncounterTrade getValidIngameTrade(PKM pkm, GameVersion gameSour
 
             if (!pkm.WasIngameTrade)
             {
-                if (pkm.HasOriginalMetLocation)
+                if (!pkm.HasOriginalMetLocation)
```
javierhimura: I'm thing this should be reverted
If a pokemon has original met location and WasIngameTrade then is impossible for the pokemon to be a be from a valid in game trade, then !pkm.WasIngameTrade, pkm.HasOriginalMetLocation return null
If the pokemon is a gen1/2/3/4 transfer WasIngameTrade will never be true because the variable is based on MetLocation whose data was lost during the transfer, but that does not mean the pokemon could not be originated from a ingame trade

wwwwwwzx: A debug thinking, I will fix soon.
since at first for sm, it goes into this part and returned

wwwwwwzx:
Also steelix is recognized as invalid now since Onix is also in the table.

javierhimura:
I think is neccesary a refactor for the get static and trade encounters to allow to return multiple encounters when more than one could be initialy possible

wwwwwwzx:
And some gift with specific OT ID, which has met location different from default, 
 will be filtered out by the wasingametrade-hasoriginallocation two if, still needs to adjust the logic here..

PKHeX/Legality/Core.cs
```diff
@@ -942,48 +937,54 @@ internal static EncounterTrade getValidIngameTrade(PKM pkm, GameVersion gameSour
 
             EncounterTrade[] table = getEncounterTradeTable(pkm);
 
-            EncounterTrade z = table?.FirstOrDefault(f => p.Any(r => r.Species == f.Species));
+            IEnumerable<EncounterTrade> poss = table?.Where(f => p.Any(r => r.Species == f.Species));
```
evandixon: I recommend making this a list, since you're calling Count() then enumerating through it.  This could have a minor performance benefit depending on whether or not Linq stores the results of the first enumeration.
```
var poss = table?.Where(f => p.Any(r => r.Species == f.Species)).ToList();
```

javierhimura:
About those gift pokemon with different data than others gift/trade the location can be specified when the      EncounterTrade is declared. The default value is just that, a default value, is not a constant, it is usefull for those halft encounters with all the same value but it a few have a different value it can be assigned in the variable

javierhimura:
I think the only thing that is yet to implement for gens 3-5 ingame trades will be create the multi language names files for every game and add the minimun level for every in-game trade (like the ones in gen 6)

PR #1008, opened 2017-03-28
Fix issue #1007 and some generation 4 encounter fixed

javierhimura:
Fix issue #1007 loading one empty item into the catch ball combo box, after that error was thrown because it was trying to select first element in a combobox without elements. 

For previous experience with .NET proyects with this same kind of code when is trying to select one element in a combo without elements sometimes just continue without exceptions just keeping selectedvalue -1and some other time throw an exception, working different from one user to another. In my case in VS in debug mode the program was not throwing exceptions.

But even if the selection is skipped the function verifiedPKM does not allow to save the pokemon because ball combo does not have a valid selected value.

Maybe is not the most elegant way to fix it, i am not very familiar with the UI code.

Also some other missing generation 4 encounters
- Trap Floor in Team Rocked HQ
- Headbutt tutor move

Fixed load raw data from generation 4 encounters. I thought time replace slot make impossible the normal slots that it replace if all 3 happens but plying with diamond i found a Ledian in route 229, encounter that bulbapedia have it but the legal analysis was marking it as illegal because its slot was removed because of the 3 time of day slots.

Also fix minimun and maximun level for DPPt Water/Fishing/Rock Smash slot, is in different order than HGSS and fix offset for time replace slots.

### DIFF
--- a/PKHeX.WinForms/MainWindow/Main.cs
+++ b/PKHeX.WinForms/MainWindow/Main.cs
@@ -1454,7 +1454,8 @@ namespace PKHeX.WinForms
             TB_SID.Text = 54321.ToString();
             int curlang = Array.IndexOf(GameInfo.lang_val, curlanguage);
             CB_Language.SelectedIndex = curlang > CB_Language.Items.Count - 1 ? 1 : curlang;
-            CB_Ball.SelectedIndex = Math.Min(0, CB_Ball.Items.Count - 1);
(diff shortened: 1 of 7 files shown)


evandixon:
How many things try to select from `CB_Ball`?  It'd probably be better to ensure whatever's selecting  doesn't call `.First()`, instead calling `.FirstOrDefault()` and handling null return values.

PR #1013, opened 2017-03-30
Safari ball and sport ball checking

javierhimura:
Remember getValidWildEncounters was implemented to filter by ball for safari zone and bug catching contest in case there were more that one possible encounter, that means to check safariball and sportball checking the first slot is enought.

getValidWildEncounters will never return an array of slot with safari and non safari encounters at the same time, nor bug-catching encounters and non bug-catching encounters

### DIFF
--- a/PKHeX/Legality/Checks.cs
+++ b/PKHeX/Legality/Checks.cs
@@ -1356,7 +1356,7 @@ namespace PKHeX.Core
 
                 if (pkm.Met_Location == 30016 && pkm.Gen7) // Poké Pelago
                     verifyBallEquals(4); // Pokeball
-                // For gen3/4 safari zones and BCC verify encounter already filter to not return
+                // For gen3/4 safari zones and BCC getValidWildEncounters already filter to not return
                 // mixed possible encounters between safari, BCC and other encounters
                 // That means is the first encounter is not safari then there is no safari encounter in the array
                 else if (3 <= pkm.GenNumber && pkm.GenNumber <= 4 && Legal.IsSafariSlot(enc.First().Type))
--- a/PKHeX/Legality/Checks.cs
+++ b/PKHeX/Legality/Checks.cs
@@ -1352,8 +1352,17 @@ namespace PKHeX.Core
             }
             if (EncounterType == typeof (EncounterSlot[]))
             {
+                EncounterSlot[] enc = EncounterMatch as EncounterSlot[];
+
                 if (pkm.Met_Location == 30016 && pkm.Gen7) // Poké Pelago
                     verifyBallEquals(4); // Pokeball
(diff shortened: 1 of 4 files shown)


javierhimura:
Actually in PkHex intially Manaphy was not in the list NoHatchFromEgg, if was added because in gen 5,6 or 7 games there is no Manaphy eggs

In fact in gen4 Manaphy only hatch from event eggs, if is included in the list a Manaphy egg from the day care coulple would appear legal.

wwwwwwzx:
also in gen4 personal table, the formcount is not recorded for altform>0. original form of giratina for example.

wwwwwwzx:
It seems the special moves were blocked by egg moves, steel wing of scizor and extreme speed of dragonite from pgl gift are not detected.

javierhimura:
Gen 4 forme count for Giratina is 0 in DP, but have value 2 in HGSS and Pt
Is correct, Giratina second form does not exits in Diamond/Pearl, was introduced in Platinum

wwwwwwzx:
gen4 personal info is using hgss, original form in hgss is illegal

javierhimura:
Actually is not illegal.
http://bulbapedia.bulbagarden.net/wiki/Timespace_orbs

Griseous Orb
HGSS 	Sinjoh Ruins (held by Giratina)

wwwwwwzx:
I mean with orb

javierhimura:
Dragonite problem is another different problem, is have nothing to do with egg moves, i debug verifymmove for that pokemon and the egg moves valid list is empty.

The problem is in getValidStaticEncounter. The function returns the first valid static encounter for that dragonite, and is dream world with special move 9. The function must be refactor to return an array of encounters when there is more than one valid encounter that match the pokemon

javierhimura:
You mean to verify if giratina form match giratina held item?

wwwwwwzx:
No I just mean the giratina from sinjoh ruins with orb is invalid now due to the incorrect  formcount recorded in hgss personal table

PKHeX/Legality/Checks.cs
```diff
@@ -791,6 +791,7 @@ private CheckResult verifyEncounterG3Transfer()
             CheckResult InvalidTransferResult = null;
             CheckResult EggResult = null;
             CheckResult G3Result = null;
+            object G3Encounter = null;
```
evandixon: Change `object` to `CheckResult`, since strong typing makes things much safer.

javierhimura: G3Encounter cant be of type CheckResult, is not to store the result, it should store the encounter to have EncounterMatch fill after the function,

EncounterMatch is declared as object because it can contain a variable of type EncounterSlot, EncounterStatic, EncounterTrade or int (for eggs)

Issue #1011, opened 2017-03-30
Legality Checker Errors (3rd, 4th and 5th gen save file)


Real96:
These pokemon are seen as not legit:

[Deoxys from FireRed](https://github.com/kwsch/PKHeX/files/880809/386.-.Deoxy.-.2F0C67684909.zip)
[Manaphy from Platinum](https://github.com/kwsch/PKHeX/files/880783/490.-.Manaphy.-.9C317942EF72.zip)
[Moltres from SoulSilver](https://github.com/kwsch/PKHeX/files/880782/146.-.Lavados.-.CCFF5838C27A.zip)
[Lugia from Emerald](https://github.com/kwsch/PKHeX/files/880829/249.-.-.4BBC7942EF72.zip)
[Latios from Emerald](https://github.com/kwsch/PKHeX/files/880830/381.-.-.FAC77942EF72.zip)
[Jirachi from Colosseum Bonus Disk --> Ruby](https://github.com/kwsch/PKHeX/files/880835/385.-.JIRACHI.-.E02C987ED636.zip)
[Arceus from entralink Black 1](https://github.com/kwsch/PKHeX/files/880855/493.-.Arceus.-.413C98221B14.zip)
[Dragonite entralink - global link event Black 1](https://github.com/kwsch/PKHeX/files/880856/149.-.Dragonite.-.71FBFFD86DF2.zip)

The 1st pokemon get all 4 moves as invalid
The 2nd pokemon get: "Invalid: Can't obtain egg for this species."
The 3rd pokemon get "Invalid: Unable to match an encounter from origin game."
The 4th and the 5th pokemon get: "Invalid: Fateful Encounter should not be checked"
The 6th pokemon get: "Invalid: Unable to match to a Mystery Gift in the database."
The 7th get: "Invalid: Hidden Ability mismatch for encounter type."
The 8th get: "Invalid Move 4: Invalid Move."

javierhimura:
- Deoxys. Done, pending merge request. Also it have another errors after transfering it to gen 4, done too
- Manaphy. Done, pending merge request.
- Moltres. Already done in the master branch
- Lugia and Latios. I also have myselft a lugia and moltres from navel rock with fatefull encounter. Maybe that means pokemon catch in the gen 3 event islands all have fatefull encounter?
- Jirachi. Events for generation 3 are not yet implemented.
[...]

PR #1023, opened 2017-04-02
Static encounter refactor to return a list for pokemon with multiple statics but different moves

javierhimura:
getValidStaticEncounters changed to return a list, the treatment of the list is similar than the mistery gift list for events encounters, it works like mistery gift encounters, the dragonite from DreamWorld in the issue #1011 now its flagged as legal

Also changed gen 1 and gen 2 encounters to take into account that the min species for an encounter can be a gen 2 preevolution, like tyroge -> hitmonlee, hitmonchan. 

When searching for static and wild encounters in gen 1 and 2 it will be used the maxSpecies of the game where the encounter is being searched, not the maxSpecies from the pkm format, the ordenation should return the min species in the evolutionary family tree, again tyrogue before himonlee or hitmonchan

Separate VC Events for GBCart Events. The event mew was being excluded in getValidStaticEncounter if AllowGBCartEra is false.

About the memory card files probably the save date is stored but in case a user have multiple savegames in the same memory card it should be possible to edit any pokémon savefile from the cart because maybe the user want to edit a different savegame that the last game he/she played.

### DIFF
--- a/PKHeX/Game/GameVersion.cs
+++ b/PKHeX/Game/GameVersion.cs
@@ -43,8 +43,9 @@
         GBCartEraOnly,
         Stadium,
         Stadium2,
-        EventsGen1,
-        EventsGen2,
+        EventsGBGen1,
+        EventsGBGen2,
+        VCEvents
     }
 
     public static class Extension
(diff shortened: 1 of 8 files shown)


kwsch:
There's currently false negatives regarding Dream World Pokémon due to all moves being stored in a single encounter. Some are tutor moves in B2W2 (and should be removed from the exclusive encounter list), but some have 2+ exclusive moves that can't be obtained at the same time (Growlithe with Body Slam AND Endure is not possible, but shows up as legal). 

With all that splitting it may be best to just turn the dreamworld pkm into a binary (maybe 4 bytes per pkm, files separated by area). The save file stores u16 move, u16 species, with some bit mingling to store gender/form iirc. Level was retrieved via a NARC file which had 1 byte per species indicating the min level. Probably isn't worth replicating that; just make it 5 bytes (species, move, 2bitflag gender/6bit form).

PR #1024, opened 2017-04-03
Fixes obedience flag to fateful encounter flag for palparking.

kamronbatman:
Only pokemon with obedience flags should have fateful encounter flags in gen 4.
This includes mew/deoxys, and for some reason eon ticket lati@s.


kamronbatman:
I forgot to update the branch to the latest master.  We can rebase or squash once approved.

PR #1027, opened 2017-04-05
Misc fixes

wwwwwwzx:
- fix some unintentionally changes while loading the binary file (location ID, nickname)
- fix around hgss Togepic egg


kwsch:
I just updated the move checking method to be ambiguous towards MysteryGift and EncounterStatic; it now loops for all possible encounters and uses their special moves if there are any.

wwwwwwzx:
Great! The type of Encountermatch is really confusing.
I tried the rest parts, they are still necessary for getting special moves of in-game gift egg.

PR #1028, opened 2017-04-06
Fix gen7 Eevee eggs checking and cleanup

wwwwwwzx:
- there is a encounterstatic entry of Eevee egg in sm, needs to mark EncounterStaticMatch as null after checking. It happens when it's not the gift Eevee egg, i.e it has valid relearn moves
- remove some unnecessary code after refactoring
- skip valid level up move when checking levelup egg moves (wasegg pokemon)
- Add PokePark Encounter
- Add Egg Pokémon Present Eggs Encounter
- Fix gen3 eggs encounter recognition

### DIFF
--- a/PKHeX/Legality/Checks.cs
+++ b/PKHeX/Legality/Checks.cs
@@ -2478,6 +2478,14 @@ namespace PKHeX.Core
         {
             CheckResult[] res = new CheckResult[4];
 
+            // Gen 1-3 could have an egg origin and a non-egg origin, check first non-egg origin
+            if (pkm.GenNumber <= 3 && !pkm.HasOriginalMetLocation && EncounterMatch != null)
+            {
+                res = parseMovesSpecialMoveset(Moves, validLevelMoves, validTMHM, validTutor);
+                if (res.All(r => r.Valid)) // moves are satisfactory
+                     return res;
+            }
+
             // Some games can have different egg movepools. Have to check all situations.
             GameVersion[] Games = { };
             switch (pkm.GenNumber)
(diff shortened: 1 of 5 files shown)


PKHeX/Legality/Checks.cs
```diff
@@ -2503,7 +2503,7 @@ private CheckResult[] parseMovesWasEggPreRelearn(int[] Moves, List<int>[] validL
                 for (int i = 0; i <= splitctr; i++)
                 {
                     var LvlupEggMoves = Legal.getBaseEggMoves(pkm, i, ver, 100)?.ToArray() ?? new int[0];
-                    var EggMoves = !pkm.WasEventEgg && !pkm.WasGiftEgg ? Legal.getEggMoves(pkm, i, ver).ToArray() : new int[0];
+                    var EggMoves = Legal.getEggMoves(pkm, i, ver).ToArray();
```
javierhimura: What conflict have you found?
I thinks this code should not be removed, egg moves are not available for event egg and gift egg, because they are generated with a fixed move set that should match the encounter exactly.

If an event egg have a move that is also a egg move from that species then the move should be added to the encounter to allow that egg to have that specific egg move but not all the other egg moves the species could learn

wwwwwwzx: Note here egg moves should be intentionally put into `parseMoves` for incompatible checking.

javierhimura:
I think with the changes gen3 pokemon will not enter to check the non-egg encounter because the wasegg condition will have preference.

Proposal Add null encounter to the list only for WasEgg pkm, and considere null encounter for potential egg encounters for the pokemon

If the pokemon WasEgg from generations 2 or 4 to 7 and the encounter list is empty then add null to check egg encounter. 

If the pokemon is from generation 3 and WasEgg add always null encounter to the list. 

Is is from generation 3 and IsEgg do like generations 2,4-7 becuse inside egg is possible to difference normal eggs from special/gift eggs

Also the parseMoves for WasEggPreRelearn should not be used for generation 3 with not null encounter match because that would be the alternate non-egg encounter from a gen 3 that could have both egg and non-egg encounter

wwwwwwzx:
Non-egg encounter for gen3 is added to the list before the wasegg encounter.
And null encounter is already add to list when `encountermatch == null`

javierhimura:
Yes, but generation 3 pokemon that could be from an egg have true in WasEgg property event if there is a possible non-egg encounter. And WasEgg have preference over the EncounterMatch. It should enter into parseMovesWasEgg only if the encounter is from a egg encounter

wwwwwwzx:
Can you give a example of non-egg encounter but have `wasegg == true`

javierhimura:
Any generation 3 pokemon that can be hatched from an egg, have met level above 5, is inside a pokeball and it is in generation 4 or more. There is no way to know if that pokemon was an egg or was from a wild/static/trade encounter, because the information that determine if a pokemon wasegg is lost when it is transfer to generation 4.

wwwwwwzx:
That means every breedable pkm transfered from gen3 can always be legal. I can't think of any wild/static/trade breedable pkm can not be obtained through breeding

javierhimura:
Then just maybe normal encounters for gen3 with wasEgg shoul be ignored for parseMoves. They are irrelevant without having any encounter moves. The same with gen 1 and 2

wwwwwwzx:
Anyway I will put it back

PR #1029, opened 2017-04-06
Honey Tree Editor for gen 4

suloku:
This is a feature that as far as I know hasn't been implemented in any save editor so far.
I'm afraid I didn't manage to keep in tune with pkhex coding style and that there's probably better ways to implement a honey tree editor, but this is the best I could manage.

I tried to keep everything in the SAV_HoneyTree.cs file to modify the minimum amount of code. The only thing that feels a little dirty is that the best way I found to access the correct offset without modifying sav4.cs was using the Pokedex offset because it had public permision access:
```
SAV.PokeDex - 0x12DC + DP_OFFSET
SAV.PokeDex - 0x1328 + PT_OFFSET
```
It works but doesn't feel quite right to do it that way.

I've been testing it for a while and I didn't find any more bugs. There's also a warning for when the user tries to set a munchlax in a non-munchlax tree (since that become an illegal pokémon).


PKHeX.WinForms/Subforms/Save Editors/Gen4/SAV_HoneyTree.cs
```diff
@@ -0,0 +1,404 @@
...
+            Burmy,
+            Cherubi,
+            Combee,
+            Heracross,
+            Munchlax,
+            Silcoon_Wurmple, //Silcoon/cascoon aren't available in Pt
+            Wurmple //Not used for Pt
+        };
+
+        private int getTreeSpecies()
```
evandixon: It might be better to move this logic to the SAV4 class

kwsch: Done
https://github.com/kwsch/PKHeX/commit/d19ab6159f6e6d9cd00c929f9e2662c19333d181

PR #1030, opened 2017-04-07
Fix and simplify gen4 encounter file reading 

wwwwwwzx:
- Change reference offset to absolute offset in order to read data independently and for troubleshooting
- Swarms species ID of dppt can be read from encounter raw file (2 x int32 starting from 0x66)
- [Swarms species ID of hgss is at 0xBE but lacking encounter type info]
- Timereplace slots of dppt only have morning and night slots, the 1st slot is just swarms
- Pokeradar slots should replace normal slots 4,5,10,11 (tested)
- While replacing slots, skip the encounters with same species ID (always because there is no such slots)

### DIFF
--- a/PKHeX/Legality/Tables4.cs
+++ b/PKHeX/Legality/Tables4.cs
@@ -570,9 +570,10 @@ namespace PKHeX.Core
         #endregion
 
         // Encounter Slots that are replaced
+        internal static readonly int[] Slot4_Swarm = {0, 1};
         internal static readonly int[] Slot4_Time = {2, 3};
         internal static readonly int[] Slot4_Sound = {2, 3, 4, 5};
-        internal static readonly int[] Slot4_Radar = {6, 7, 10, 11};
+        internal static readonly int[] Slot4_Radar = {4, 5, 10, 11};
         internal static readonly int[] Slot4_Dual = {8, 9};
         #region Alt Slots
         internal static readonly int[] SafariZoneLocation_4 =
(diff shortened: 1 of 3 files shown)


PKHeX/Legality/Tables4.cs
```diff
@@ -570,9 +570,10 @@ public static partial class Legal
         #endregion
 
         // Encounter Slots that are replaced
+        internal static readonly int[] Slot4_Swarm = {0, 1};
         internal static readonly int[] Slot4_Time = {2, 3};
         internal static readonly int[] Slot4_Sound = {2, 3, 4, 5};
-        internal static readonly int[] Slot4_Radar = {6, 7, 10, 11};
+        internal static readonly int[] Slot4_Radar = {4, 5, 10, 11};
```
kwsch: Radar slots are 6 & 7 tho
https://sites.google.com/site/pokemonslots/gen-iv/diamond

wwwwwwzx: But when I read from the data, I can see when there is no pokeradar encounter, the species ID are identical to slots 4,5. Also if I change slots to 4,5 the save file of RoC will be all clear (legal) now

wwwwwwzx:
I'm using Matlab for binary reading. [Here](https://gist.github.com/wwwwwwzx/2defb8ecf61163201d727fa847e33288) is my testing code

wwwwwwzx:
The order of google site and of encounter file are the same. That's weird
The encounter data of Valor Lakefront (Pearl): The number in the first row are levels, the others are species ID.

the level stats from bulbapedia

If Nidorino(33) replace (6,7) it will only level at 22
if replace (4,5) it will have 21 or 22

kwsch:
according to http://www.geocities.jp/bt_vermeille/remodel/pokemon_pl/

```
	dword	入れ替えＡ７(ポケトレ、データ５と入替)　種類
	dword	入れ替えＡ８(ポケトレ、データ６と入替)　種類
	dword	入れ替えＡ９(ポケトレ、データ１１と入替)　種類
	dword	入れ替えＡ１０(ポケトレ、データ１２と入替)　種類
```
slots are not listed as 0 index (instead 1,2,3,4).
ポケトレ = Poké Radar

Guess the RNG tables are wrong (who uses radar for RNG anyway)

PR #1039, opened 2017-04-09
Poketch DotArtEdit

pokecal:
drag&drop will import file, click will edit directly.


pokecal:
found bug?, I will make another PullRequest.

PR #1040, opened 2017-04-09
Poketch DotArtEdit

pokecal:
drag&drop will import file, click will edit directly.


pokecal:
Thanks, I will learn more about git commands and use it.

PR #1050, opened 2017-04-10
E-reader Berry

javierhimura:
I change the code to check the legallity of the e-reader held item.

Enigma berry was releashed but only as an e-berry replacer, that means to be legal the held-item it must be shown in game as an e-reader berry, not as an enigma berry. In fact, in FR, LG and Emerald games the item is unreleashed.

Now the programa load from the savegame the e-reader name if there is a valid e-reader loaded in the save.

If there is a e-berry the items list show the e-berry name instead of enigma berry, just like the games do.

Also the legality check mark the default enigma berry as unreleashed but allow the item if is a e-berry, checking if the berry match the name of one of the available e-berries for the save language.


kwsch:
I can revise this later, but SAV3 and SaveUtil functions should not interact with legality checking settings. Like determining whether or not GBEra should be considered, the E-Reader data config should be set when loading a save file (method: `openSAV` in `Main.cs`).

Has to be segregated in the event PKHeX.Core dll is reused in another application where multiple saves may be in use.

PKHeX/Game/GameInfo.cs
```diff
@@ -254,7 +254,16 @@ public string[] getItemStrings(int generation, GameVersion game)
                             case GameVersion.XD:
                                 return g3xditems;
                             default:
-                                return g3items;
+                                {
```
evandixon: Nit pick: you don't need curly braces here.  A switch statement is a glorified set of go-to statements, which aren't themselves blocks.

javierhimura:
Thanks, i change the comunication to legality class to main.cs
Should i change also the code of GameInfo.cs to load the berry display name without using legality?

PR #1051, opened 2017-04-11
Varios legal analysis improvements and fixes

javierhimura:
- Ignore Shedinja evolution move for Shedinja encounters. There are some Shedinja events available

- Remove unused Sevii Islands locations from valid fr lg met locations

- Do not allow Phione and Manaphy to inherit level up moves from level 2 to 100

- Do not allow hidden ability for generation 3-4 pokemon

- Do not allow egg hatch cycles greater than the species hatch cycles or the encounter hatch cycles for the two ORAS gift eggs.

- Verify abilties for generation 5 events, is the same verification from gen 6-7 but ability capsule is not valid for format = 5

- Fix gender validation. Some gen 3 pokemon with a male-female value of 127 was given errors, because the calc in PKX.getGender and PKM.getGenderIsValid was different

- Added generation 3 ability verification. Some pokemon has one ability in generation 3 and 2 in generation 4. That means if they evolve in generation 4-5 games the ability will change and then it will match the PID. The legallity change based in that evolution.

The validation uses the following rules.
- If the pokemon have one ability in generation 4 then do not need to check anything

- If the pokemon has two abilities in generation 3 then it must match the PID except for Colloseum encounters

- If the pokemon has one ability in generation 3 and can be proven it have NOT evolved in generation 4 or 5 then only the generation 3 ability is valid, without matching PID. For example a gen 3 Porygon with Download is illegal

- If the pokemon has one ability in generation 3 and can be proven it have evolved in generation 4 or 5 then it must have the ability that match the PID. For example a generation 3 Porygon-Z

- In other cases where both options are valid the pokemon could have the generation 3 ability without matching PID or it could have the generation 4 second ability, but in that second case ability must match PID. For example a generation 3 Porygon-2, there is no way to tell if it has evolved in gen3 games or gen4-5 games. Porygon-2 with Trace is legal with any PID, Porygon-2 with Download is legal if PID match ability.


javierhimura:
I just found bulbasaur moves from firered level up table (not charmaander like i say in the commit) was wrong, with 67 different moves. I modify the file to have the same learnset as the file from leaf green.
Now the only pokemon with different movesets between FR and LG are Deoxys and Dugtrio (the second one have different order between the initial moves, but are the same moves)

PKHeX/Legality/Checks.cs
```diff
@@ -1356,6 +1374,72 @@ private void verifyAbility()
...
+                // gen3Species will be zero for pokemon with illegal gen 3 encounters, like Infernape with gen 3 "origin"
+                // Do not check for gen 3 pokemon that has evolved into gen 4 species, 
+                // those have evolved in generation 4 or 5 and ability must match PID, not need to check gen 3 data
+                var gen3Species = EvoChainsAllGens[3].LastOrDefault()?.Species ?? 0;
+                if(gen3Species > 0)
+                    AbilityMatchPID = verifyAbilityGen3Transfer(abilities, abilval, gen3Species, abilities_count);
+            }
+            
+            // Gen 4,5 pokemon or gen 3 pokemon evolved in gen 4,5 games, ability must match PID
+            if (AbilityMatchPID && abilities_count == 2 && pkm.AbilityNumber != 1 << abilval)
```
evandixon: `abilities_count == 2` is redundant since AbilityMatchPID (which is the same thing) is already true.

PR #1052, opened 2017-04-11
Fixes gen 5 force-shiny generation.

kamronbatman:
Identified an issue with force-ability and force-shiny WCs where the PID was off.
For example:
Original PID was A3134B185
Ability Forced to Ability 1
Forced to be Shiny

PID should be 03720085.  PKHex generated it as 03720084.

There MIGHT be an issue with force ability HA. I have theories on it, but would need to disassemble or generate them to see what actually happens.

### DIFF
--- a/PKHeX/MysteryGifts/PGF.cs
+++ b/PKHeX/MysteryGifts/PGF.cs
@@ -14,7 +14,7 @@ namespace PKHeX.Core
             if (data == null) Data = new byte[Size];
             else Data = (byte[])data.Clone();
         }
-        
+
         public ushort TID { get { return BitConverter.ToUInt16(Data, 0x00); } set { BitConverter.GetBytes(value).CopyTo(Data, 0x00); } }
         public ushort SID { get { return BitConverter.ToUInt16(Data, 0x02); } set { BitConverter.GetBytes(value).CopyTo(Data, 0x02); } }
         public int OriginGame { get { return Data[0x04]; } set { Data[0x04] = (byte)value; } }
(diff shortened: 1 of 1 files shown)


kwsch:
Reference [RNG Reporter](https://github.com/Slashmolder/RNGReporter/blob/e9128b166c8ed005fabbc8939b105c77fa838723/RNGReporter/Objects/FrameGenerator.cs#L4317-L4551); can confirm yours is correct, does the same as ForceShiny. Also note this is a good ref for eventual legality PID method determination.

It doesn't look like RNG Reporter does anything with ability bits, but my suspicion is that the game clears the ability bit from the PID and only sets it back if ability1 is to be generated. It wouldn't have any impact on the shininess as the bit is shifted out. So, the proposed implementation if-else would be inverted:

```
if (av == 1)
    pk.PID |= 0x10000; // set ability bit
else 
    pk.PID &= 0xFFFEFFFF; // clear ability bit
```

kamronbatman:
I didn't do it like that because of HA.  I am not sure how HA is handled, but I assumed av ==2 would be the same as av == 0?

Also Kirzi generated lots of them and confirmed the bug.  So RNG Reporter is most likely wrong too.

kwsch:
Only way to find out is to let the game generate some HA pkm; my vague recollection of how it works may be incorrect ;)

For RNG Reporter the accuracy of the PID was inconsequential as the ability bit doesn't impact shininess, and nobody cared to RNG a gift for characteristics.

kamronbatman:
Kirzi is doing that now actually.  So for now this code is correct, minus HA.

Kirzi:
https://pastebin.com/eYXkEi92

More later maybe.

kamronbatman:
So according to the HA, it is treated like ability 1.  I'll make the adjustments.

kamronbatman:
@kwsch Updated to the if statement you proposed.  That handles the forced HA which Kirzi generated.  Should be ready for merging now.

kwsch:
https://github.com/kwsch/PKHeX/commit/30706835adf51c45d41e4481fc7698217fbc1b23

Moved the ability forcing to the end so that the antishiny and regular abilities are forced the same way as documented in this issue.

PR #1053, opened 2017-04-12
Fix abilties for pokemon with second ability with zero

javierhimura:
A simple fix, there are pokemon with the second ability 0 in the personal tables like the Kanto Starter in personal table for Emerald and Hearth Gold/Soul Silver.

The same table also have other pokemon with the second ability equals to the first ability.

### DIFF
--- a/PKHeX/Legality/Checks.cs
+++ b/PKHeX/Legality/Checks.cs
@@ -1281,7 +1281,7 @@ namespace PKHeX.Core
                 return;
             }
 
-            var abilities_count = abilities.Distinct().Count();
+            var abilities_count = abilities.Where(a => a != 0).Distinct().Count();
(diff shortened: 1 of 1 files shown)


kwsch:
`abilities` is provided [after screening for 0](https://github.com/javierhimura/PKHeX/blob/21ae8f6766bf71b47f58fca55faef6b1e7ff6070/PKHeX/Legality/Checks.cs#L1236-L1237)

```
            int[] abilities = pkm.PersonalInfo.Abilities;
            if (abilities[1] == 0)
                abilities[1] = abilities[0];
```

so no ability should be zero unless it's the missingno entry, thus no need to Where() check.

javierhimura:
I didnt notice that part, only the raw file itself.  I cancel the pull request, thank you

PR #1056, opened 2017-04-14
Gen3/5 Ability Check Fixes

wwwwwwzx:
- Separate AbilityNumber(from PID) and Ability(from IV32) of PK3
- Fix gen5 ability checks (b2w2 personal table contains HA, the ability_count won't work)
- Consider Ability Capsule when compare encounter ability
- Use bool? to mark 3 ability states
- Simplify and avoid duplicate invalid messages (PID mismatch < AbilityNumber mismatch)
- Fix gen3 in-game trade ability check


javierhimura:
Im not sure if this change will work with pokemon from eggs

EvoChainsAllGens = Legal.getEvolutionChainsAllGens(pkm, EncounterOriginalGB ?? EncounterMatch ?? pkm.Species);

getEvolutionChainsAllGens call the function getEvolutionChain and this function use the encounter to determine the min species from the encounter.

If the encounter is the current pokemon integer species when then any pre evolutions would be ignored, maybe that will be problematic with pokemon with egg origin, because preevolutions will not be included in the evolution chain. Maybe change pkm.Species and instead use Legal.getBaseEggSpecies

PKHeX/PKM/PK3.cs
```diff
@@ -116,7 +116,7 @@ public PK3(byte[] decryptedData = null, string ident = null)
         public override int IV_SPA { get { return (int)(IV32 >> 20) & 0x1F; } set { IV32 = (uint)((IV32 & ~(0x1F << 20)) | (uint)((value > 31 ? 31 : value) << 20)); } }
         public override int IV_SPD { get { return (int)(IV32 >> 25) & 0x1F; } set { IV32 = (uint)((IV32 & ~(0x1F << 25)) | (uint)((value > 31 ? 31 : value) << 25)); } }
         public override bool IsEgg { get { return ((IV32 >> 30) & 1) == 1; } set { IV32 = (uint)((IV32 & ~0x40000000) | (uint)(value ? 0x40000000 : 0)); } }
-        public override int AbilityNumber { get { return 1 << (int)((IV32 >> 31) & 1); } set { IV32 = (IV32 & 0x7FFFFFFF) | (value > 1 ? 0x80000000 : 0); } }
+        public byte AbilityBit { get { return (byte)(IV32 >> 31); } set { IV32 = (IV32 & 0x7FFFFFFF) | (uint)(value << 31); } }
```
evandixon: This property is a single bit, right?  Why not make it a boolean?
```
public bool AbilityBit { get { return (IV32 >> 31) == 1; } set { IV32 = (IV32 & 0x7FFFFFFF) | (uint)(value ? 1 : 0 << 31); } }
```
Line 32-33 of MainPK3.cs would then become the following:
```
bool abil = pk3.AbilityBit;
CB_Ability.SelectedIndex = abil && CB_Ability.Items.Count == 0 ? 0 : 1;
```

PKHeX/PKM/PK3.cs
```diff
@@ -116,7 +116,7 @@ public PK3(byte[] decryptedData = null, string ident = null)
         public override int IV_SPA { get { return (int)(IV32 >> 20) & 0x1F; } set { IV32 = (uint)((IV32 & ~(0x1F << 20)) | (uint)((value > 31 ? 31 : value) << 20)); } }
         public override int IV_SPD { get { return (int)(IV32 >> 25) & 0x1F; } set { IV32 = (uint)((IV32 & ~(0x1F << 25)) | (uint)((value > 31 ? 31 : value) << 25)); } }
         public override bool IsEgg { get { return ((IV32 >> 30) & 1) == 1; } set { IV32 = (uint)((IV32 & ~0x40000000) | (uint)(value ? 0x40000000 : 0)); } }
-        public override int AbilityNumber { get { return 1 << (int)((IV32 >> 31) & 1); } set { IV32 = (IV32 & 0x7FFFFFFF) | (value > 1 ? 0x80000000 : 0); } }
```
kwsch: Removing the AbilityNumber get/set breaks the conversion from CK3/XK3 -> PK3, as well as any backwards conversion (pk2pk uses the property-reflection transfer to go 4->3).

Just restore this deleted line :)

wwwwwwzx: I put the AbilityNumber below as a generated Attributes，But no set function, wonder if It's okay to set the topbit instead of PID. If not to separate them, the pid ability check of pk3 will be always valid.

PR #1064, opened 2017-04-14
Fix Bug Contest Butterfree error reported in issue #1011

javierhimura:
There is in the national park a butterfree encounter as a normal wild pokemon and another with the same levels in the bug contest, that means even in format 4 it could exits more than one type encounters in the list, it should be filtered

### DIFF
--- a/PKHeX/Legality/Core.cs
+++ b/PKHeX/Legality/Core.cs
@@ -1148,10 +1148,10 @@ namespace PKHeX.Core
             foreach (var area in getEncounterAreas(pkm, gameSource))
                 s.AddRange(getValidEncounterSlots(pkm, area, DexNav: pkm.AO, gameSource: gameSource));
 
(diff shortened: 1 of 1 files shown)


PKHeX/Legality/Core.cs
```diff
@@ -1148,10 +1148,10 @@ internal static EncounterSlot[] getValidWildEncounters(PKM pkm, GameVersion game
             foreach (var area in getEncounterAreas(pkm, gameSource))
                 s.AddRange(getValidEncounterSlots(pkm, area, DexNav: pkm.AO, gameSource: gameSource));
 
-            if (s.Count <= 1 || 3 > pkm.GenNumber || pkm.GenNumber > 4 || pkm.HasOriginalMetLocation)
+            if (s.Count <= 1 || 3 > pkm.GenNumber || pkm.GenNumber > 4 || (pkm.Gen3 && pkm.HasOriginalMetLocation))
```
kwsch: Shouldn't this be pkm.Gen4 not Gen3?

PR #1068, opened 2017-04-16
adds for Gen3

pokecal:
All address is JapaneseVer.ROM(Em,FR), also need check with InternationalVer.

### DIFF
--- a/PKHeX.WinForms/Subforms/Save Editors/Gen3/SAV_Misc3.cs
+++ b/PKHeX.WinForms/Subforms/Save Editors/Gen3/SAV_Misc3.cs
@@ -19,9 +19,15 @@ namespace PKHeX.WinForms
                 tabControl1.Controls.Remove(TAB_Joyful);
 
             if (SAV.E)
+            {
                 readFerry();
+                readBF();
+            }
             else
+            {
                 tabControl1.Controls.Remove(TAB_Ferry);
+                tabControl1.Controls.Remove(TAB_BF);
+            }
 
             if (SAV.FRLG)
                 TB_OTName.Text = PKX.getString3(SAV.Data, SAV.getBlockOffset(4) + 0xBCC, 8, SAV.Japanese);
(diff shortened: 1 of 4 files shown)


pokecal:
@ricksee I have Pearl(just HOF) and HG(newgame), no Pt. I started Pearl to the Tower, but play HG if you want. (finally both games will begins to the Tower)

ricksee:
@pokecal HS uses the same Battle Frontier as Pt, so I assume it is the same code.

PKHeX.WinForms/Subforms/Save Editors/Gen3/SAV_Misc3.cs
```diff
@@ -187,5 +195,188 @@ private void saveFerry()
...
+            else if (BattleType >= BFT[BFF[Facility][1]].Length) return;
+
+            int RBi = -1;
+            for (int i = 0, j = 0; i < StatRBA.Length; i++)
+            {
+                if (!StatRBA[i].Checked) continue;
+                if (++j > 1) return;
+                RBi = i;
+            }
+            if (RBi < 0) return;
```
evandixon: This loop could be eliminated by using Linq. Something like this:
RBi = StatRBF.FirstOrDefault(x => x.Checked) ?? -1;

PKHeX.WinForms/Subforms/Save Editors/SAV_SimpleTrainer.cs
```diff
@@ -203,6 +208,12 @@ private void B_Save_Click(object sender, EventArgs e)
                 sav2.TextSpeed = CB_TextSpeed.SelectedIndex;
             }
 
+            if (SAV is SAV3)
+            {
+                SAV3 sav3 = (SAV3)SAV;
```
evandixon: Use "SAV as SAV3", since it's a little faster.

pokecal:
@evandixon I imitate others, methods that I have never used are weak points. The advice is very helpful to me.
@ricksee If you can't wait for controls (this weekend is my goal), please use the following data:
BP 0x5BB8, SilverTowerPrint 0xE7E+02h,
TowerSingleCurrent 0x5264, TowerSingleContinuable 0x5BC1+01h,
ContinueTimes 0x5BC2 (e.g. Current=42 -> ContinueTimes=6)

PR #1070, opened 2017-04-17
Removes items when transferred from 6 to 7

kamronbatman:
Items cannot be banked from gen 6 to 7.

### DIFF
--- a/PKHeX/PKM/PK6.cs
+++ b/PKHeX/PKM/PK6.cs
@@ -591,6 +591,7 @@ namespace PKHeX.Core
             for (var i = 0xE4; i < 0xE8; i++) pk7.Data[i] = 0; /* Unused. */
             pk7.Data[0x72] &= 0xFC; /* Clear lower two bits of Super training flags. */
             pk7.Data[0xDE] = 0; /* Gen IV encounter type. */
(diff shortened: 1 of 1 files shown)


kwsch:
Are there any held items released in gen6 that can't ever be obtained in gen7?

It's better to be nice to the user to keep held items if they exist in the destination game. Makes it easier to move old battle sets to future generations and have them ready for battle.

kamronbatman:
While that might be a good 'feature,' it makes drag-drop mystery gifts a pain in the ass.  Maybe being bank accurate should only apply to mystery gifts?  And there probably are held items which are exclusive to gen 6.  I can find out if that makes a difference?

kwsch:
Key items can't be held (unless hacked on); and mega stones depend on their released state.
Both are flagged by the legality checker; it's better to keep items when transferring so that the other items aren't removed.

kamronbatman:
The mega stones are not flagged as illegal when I pull them in.  For example a starter with venasaurite is not flagged as illegal.  I think the check for comparing to the MG overrides.

kwsch:
You can get [Venusaurite from Red at the Battle Tree](http://www.serebii.net/pokearth/alola/battletree.shtml) :)

PR #1083, opened 2017-04-22
Special Eggs improvement and Generation 4 Encounter Type legal analysis

javierhimura:
I added legal analysis for generation 4 encounter type.
Not every static encounters have yet defined the correct encounter type, i added only the ones i know the correct value, that is why i added by default EncounterType.Any for those static encounters with their value not yet implemented, it does not means they can have any encounter type

Also added two more unreleased encounter to generation 4, DP Newmoon Island Darkrai and Jhoto Route 45 Surfing (water tiles are unreachable). Also DP unreleased event locations have been removed from DP valid met locations for egg hatching, they are only valid in Platinum.

About the special eggs for generation 3 i changed the encounter order used in parseMoves to check first the normal egg encounter and after that check for the special egg encounter if the pokemon knows any moves from the special egg. With that the error output will be more complete, because special eggs check for normal egg moves but normal eggs does not check special egg moves.


javierhimura:
I added fix for the pokemon in format 5 from issue 1011.
Those pokemon have both static encounters and wild encounters from generation 4, but static encounters only allow one type of ball, i change the analysis to ignore the static encounter if the ball does not match and there is a valid wild encounter

PKHeX/Legality/Checks.cs
```diff
@@ -929,7 +973,35 @@ private CheckResult verifyEncounterG3Transfer()
         private CheckResult verifyEncounterG4Transfer()
         {
             CheckResult Gen4Result = null;
-            
+            CheckResult Gen4WildResult = null;
+            object WildEncounter = null;
```
evandixon: Probably best to use as specific of a type as you can.

javierhimura: I thought first it was not possile because EncounterMatch is an object that can be many types, but WildEncounter only store the result of getWildEncounters that have EncounterSlot[] type, i change it to that type

PR #1094, opened 2017-04-25
Fix error in transfer legality G4.

javierhimura:
Transfer legality is not checked if the pokemon is a valid static encounter

Also an Entei with invalid encounter is returning null as check encounter making the analysis to throw an exception, reported in the forum. 

Changed to do not return checkresult until the end to add invalid values if Gen4Result is null.


javierhimura:
Also i fixed the analysis to ignore static encounters if pokemon is inside a safari ball or inside a sports ball, to avoid pokemon like safari Voltorb to be reported as illegal because the static encounters take preference. 

Those balls are not used on static encounters after all.

Issue #1095, opened 2017-04-25
Generation 4 Encounter Type legality checks


PrometheusG:
Hi again,
Testing PKHeX Build 2928:

I discovered some "Encounter Type does not match encounter" errors in my Pearl Game
Most of them have Encounter Type "Wild Encounter Swarm" and some are Stark Mountain Encounters, Maniac Tunnel, Mt. Coronet, Valley Windworks Drifloon(every Friday outside) etc.

These are the pk4 files from Pearl (containing and some pokemon from Platinum):
[Pearl pk4 files.zip](https://github.com/kwsch/PKHeX/files/955854/Pearl.pk4.files.zip)

Platinum Save Extra:
I am also sending all the fossil pokemon from Platinum that show the invalid error:
"Unable to match an encounter from origin game" when the encounter type is (correctly) set as "Starter/Fossil/Gift (Pt/DP Trio)".
I found out that changing the encounter type to "Starter/Fossil/Gift (DP) makes them legal again when that should not happen in my opinion.
(I have leveled them up with rare candies to lvl 40, that is why they have 0 EV)

These are the pk4 fossil pokemon from platinum:
[Platinum pk4 files.zip](https://github.com/kwsch/PKHeX/files/955855/Platinum.pk4.files.zip)

Thank you

PrometheusG:
Hi again,

Just a small update using PKHeX build 2947
I know that you are working on Mt. Coronet  and Stark Mountain Encounters ( I read what you wrote on #1096 pull request), I just added some Great Marsh encounters I found that have encounter type as Surfing/Fishing.

Overall, the new pokemon are the ones from Great Marsh and the extreme speed Dratini from Dragon's Den which was given as a gift from the Elder, (I think TheSLAYER already posted about this on [projectpokemon](https://projectpokemon.org/forums/forums/topic/40529-legality-bugs/?do=findComment&comment=213321))

On Pearl gen 4 sav file:
Is just **Numel** and **Abomasnow** that you already have them downloaded from the Pearl pk4 files.zip I uploaded above

_On Platinum gen 4 sav file:_
[Platinum.zip](https://github.com/kwsch/PKHeX/files/961739/Platinum.zip)

These are 5 new, just take a look at the Great Marsh pokemon files

**Skarmory**: "Encounter type does not match encounter" - From Stark Mountain, Tall Grass
[...]

PR #1096, opened 2017-04-25
Fix issue #1095 and gen 4 egg encounter issue from forums

javierhimura:
Fix issue with egg pokemon reported in the forums, gen 4 transfer analysis was returning wild encounters for a pokemon that was an egg

Added encounter type to swarms and pokeradar safari slots

Split DP and Pt fossils, they have different encounter type

Change Drifloon and Spiritomb to encounter static without type, that stationary pokemon are not in a tail grass tile

Added Encounter Type to Burned Tower Suicune 

Added Maniac Tunnel to cave Location

### DIFF
--- a/PKHeX/Legality/Checks.cs
+++ b/PKHeX/Legality/Checks.cs
@@ -1015,7 +1015,7 @@ namespace PKHeX.Core
                     Gen4Result = result;
             }
 
-            if (Gen4Result == null && null != (EncounterMatch = Legal.getValidWildEncounters(pkm)))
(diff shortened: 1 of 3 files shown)


javierhimura:
Not every encounter reported in issue #1095 is fixed, Stark Mountain and Mt.Coronet have both cave and tail grass encounters, i need to check in the what raw encounter area for those locations are for the outside zones to have those as grass instead of cave. Just like Ruins of Alph in HGSS

PR #1099, opened 2017-04-26
Generation 1 and 2 legal Improvements

javierhimura:
Refactor parseMovesForEncounter to gather valid moves for species encounter, some Pokemon can have valid encounters with different source species from the encounter, the valid moves change if the encounter species change because some preevolutions moves are illegal if pokemon caught already evolved.

Example, generation 1 Pikachu that can have a RBY Pikachu encounter and GSC Pichu encounter, the valid moves for the first encounters should not have any Pichu exclusive evolution moves

Also assign the encounter match from gb when parsing moves like the variable Encounter Match, to store the encounter that is valid for the pokemon moves instead the first encounter.

Store the species encounter, this will be needed to check if the evolution is valid for species that evolve leveling with a given learned move

Add Tradeback Status to the pokemon, this variable for generations 1 and 2 use the catch rate data to determine if trade between generations 1 and 2 was possible.
If analysis is for VC games tradeback will have value NotTradeback for every gen 1 pokemon, but for cart saves some pokemon can be determine that have been never traded to generation 2, 

If catch rate match species catch rate but do not match a valid generation 2 held item that means the pokemon habe been never traded to generation 2 games, that allow to discart encounters and moves from generation 2.

If is not tradeback catch rate is used to filter encounters, catch rate determine in what species was captured the pokemon discarting some preevolutions species and moves.

Add  the option for generation 1 cart save analysis to check legal status not allowing generation 2 games, like VC games but with Stadium allowed, this are the generation 1 non tradeback rules from Smogon

Also change evolution chains to include generation 2 preevolutions from gen 1 pokemon if tradeback was possible, it is needed to avoid parsemoves to detect pokemon like Hitmonchan with Tyrogue level up moves as illegal. If the moves is only possible with a generation 1 evolution chains the previous change in getvalidmovesencounter will update the evolution chains to remove Tyrogue.

Check legal values of generation 1 type and catch rate

Replace pokemon catch rate after changing pokemon species always if the pokemon was not tradeback from generation 2, the catch rate will be keep unchanged only if it can be a held item and do not match species catch rate (default item)

Also if catch rate is changed use base species catch rate to avoid legal errors because some evolutions chatch rate does not allow preevolutions moves and some are not even possible in wild encounters.


PKHeX/Legality/Analysis.cs
```diff
@@ -212,6 +214,51 @@ private void updateEncounterChain()
...
         }
+        private void UpdateTradebackG12()
+        {
+            if (pkm.Format == 1)
+            {
+                if (!Legal.AllowGen1Tradeback)
+                    pkm.TradebackStatus = TradebackType.Gen1_NotTradeback;
+                else
+                {
+                    var catch_rate = ((PK1)pkm).Catch_Rate;
```
evandixon: Use "pkm as PK1", since it's marginally faster.

PKHeX/Legality/Structures/ValidEncounterMoves.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace PKHeX.Core
+{
+    public class ValidEncounterMoves
+    {
+        public int EncounterSpecies;
```
evandixon: It probably won't matter too much here, but properties are safer than variables. "public int EncounterSpecies { get; set; }"

PKHeX/Legality/Checks.cs
```diff
@@ -2525,63 +2574,129 @@ private CheckResult[] verifyMoves(GameVersion game = GameVersion.Any)
 
             return res;
         }
-        private void UptateGen1LevelUpMoves(ref List<int>[] validLevelMoves, List<int> DefaultMoves, int generation)
+        private void UptateGen1LevelUpMoves(ref ValidEncounterMoves EncounterMoves, int defaultLvlG1, int generation)
```
evandixon: I don't think you need the `ref` anymore, since you're now passing a reference to an instance of the ValidEncounterMoves class. The `ref` keyword makes the parameter a reference to the reference to the class.

PR #1105, opened 2017-05-01
Encounter Type fix and detection of pokemon that should have evolve on trade

javierhimura:
For every encounter trade:

Check pokemon that evolve on trade that have an encounter trade, except one that is traded with an everstone all of this must been evolved.

For every generation 1:

Check outsider pokemon that evolve on trade in generation 1 and have not been evolved, only if the encounter is the second evolution stage, it will only return illegal if filter by catch rate.

Mark as WasTradeback pokemon from one generation that only have a valid encounter in the other GB generation and pokemon with moves exclusive from the other GB generation.

Add catch rates from pokemon stadium and 3 trades with special catch rate (krabby and pokemon that evolve on trade)

And for encounter types in generation 4:

Headbutt encounter type completed. The encounter type depends on the battle background and that depend on the tile where the encounters is trigger. Heabutt have a builing encounter if the location is a city, a cave encounter for caves (like illex forest) and grass encounter for grass in routes and none in a non-combat tile in routes. In every location if the tree is headbutt from the water it will be a surfing encounter (Yes, surfing Heracross is legal).

I made a list of locations where there are trees near water, trees near grass and a list of city locations. Some locations could have more that one encounter type valid from headbutts, any route have trees in non-combat tiles but most of them have also trees near grass, other trees near water, and others have the 3 options. The same with caves and cities, some have water near trees.

I will need to make a similar test with rock smash encounters and honey trees, i think those also can have multiple encounter types.

And to fix Issue #1095 it is now included what raw files for MtCoronet and Stark Mountain is grass instead cave. The same for Mt. Silver Cave.

And finally surfing safari changed to surfing encounter type, those encounters in DPPt use the same battle background as normal surfing fishing encounters instead greath marsh encounter type.


PKHeX/Legality/Checks.cs
```diff
@@ -782,7 +803,7 @@ private CheckResult verifyEncounterG12()
                 pkm.WasEgg = true;
                 return verifyEncounterEgg();
             }
-            EncounterMatch = EncounterOriginalGB = EncountersGBMatch.FirstOrDefault()?.Encounter;
+            EncounterMatch = EncounterOriginalGB = EncountersGBMatch?.FirstOrDefault()?.Encounter;
```
kwsch: EncountersGBMatch is already checked for null at the top of the method; null check is unnecessary at this point as nothing changes what EncountersGBMatch is

PR #1109, opened 2017-05-03
add Gen3(Em) Legendary controls

pokecal:
Tested with JapaneseVer.
In game, Lati(SouthernIs.) check flags not only defeated/caught,
the FerryFlag(SouthernIs. accessability) is checked.
But the code I wrote does not consider this checking.


RGreywind:
Can u add edition for 4 and 5 gens? edit the pokemons directly from the cardtrige?

pokecal:
Special mons handling (e.g. Pearl Shaymin) is unknown.(takes long time, or postpone, not decided)
All other elements(except defeat/caught flag) are postponed.
I use Japanese carts, Pearl, HG, W, B2.
I will process from Pearl, but I accept designation of priority.

pokecal:
Pearl reserching have finished, and I think that all Gen can use same GUI layout, so I will make new subForm.

PR #1110, opened 2017-05-08
Update README.md screenshot + Hoopa forme fix

sora10pls:
To be consistent with PP.org's thread.

Edit: only downside to this screenshot is that it's relatively outdated.

Screenshot from PP.org:
image

Screenshot of latest commit:
image


kwsch:
Gotta show off all the fancy sprites and a more interesting pkm than rufflet :)

sora10pls:
Rowlet ;)

And yeah I was debating on it but decided against it for whatever reason lmao, oh well. Good to see it did get updated!

PR #1122, opened 2017-05-12
Unban Heracronite and Houndoominite

sora10pls:
Released via international serial code "AZUL"
image


sora10pls:
P.S. Thanks for the info on updating VS, got 2017 up and running a little while ago :)

PR #1136, opened 2017-05-16
Add option to load JKSV folder

sora10pls:
Actually let me close this and submit a new pull request in a few minutes. VS totally borked my commit

PR #1137, opened 2017-05-16
Add OpenJKSV function (revised)

sora10pls:
A revised approach compared to my previous (and now closed) pull request.


kwsch:
Will eventually close this PR without merging for reasons discussed in #1136, alternate (more flexible/friendly) implementation is preferred.

sora10pls:
That's fine, glad I could at least bring it to your attention!

PR #1139, opened 2017-05-17
Fixes friendship with redeemed wondercards.

kamronbatman:
Fixes an issue where redeemed WC6/WC7's have 0 friendship.  According to tests, this should be base friendship.


kamronbatman:
Per conversation, will be fixed to also include a fix for egg hatch cycles.

PR #1170, opened 2017-05-28
Mark OT Female trainer from Gamecube and VC pokemon as invalid.

javierhimura:
A small check for ot gender in GC and GB pokemon.

Gamecube games and generation 1 games do not have female player character.

Also generation 2 is not checked because if the pokemon is in format 2 and have met location data stored that means it was caught in crystal where female player is allowed, checks are not included for generation VC2 pokemon, we do not know how the bank will transfer ot gender of crystal pokemon if VC2 games are released


sora10pls:
"OT Trainer" in the legality strings should just be "OT".

Original Trainer Trainer ;)

PR #1184, opened 2017-06-02
Fix Link Gift Check if not Gen 6

ReignOfComputer:
Prevents NullReferenceException if location is set in other Gen. Reference: https://projectpokemon.org/forums/forums/topic/40749-beautifly-and-dustox-re-roll-pidencryption-pkhex-becomes-unresponsive/

### DIFF
--- a/PKHeX.Core/Legality/Encounters/EncounterGenerator.cs
+++ b/PKHeX.Core/Legality/Encounters/EncounterGenerator.cs
@@ -129,7 +129,7 @@ namespace PKHeX.Core
         private static IEnumerable<IEncounterable> GenerateRawEncounters(PKM pkm)
         {
             int ctr = 0;
-            if (pkm.WasLink)
+            if (pkm.WasLink && pkm.GenNumber == 6)
             {
                 foreach (var z in getValidLinkGifts(pkm))
                 { yield return z; ++ctr; }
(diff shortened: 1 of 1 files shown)


PKHeX.Core/Legality/Encounters/EncounterGenerator.cs
```diff
@@ -129,7 +129,7 @@ private static IEnumerable<IEncounterable> GenerateFilteredEncountersPIDIV(PKM p
         private static IEnumerable<IEncounterable> GenerateRawEncounters(PKM pkm)
         {
             int ctr = 0;
-            if (pkm.WasLink)
+            if (pkm.WasLink && pkm.GenNumber == 6)
             {
                 foreach (var z in getValidLinkGifts(pkm))
```
evandixon: I recommend simply handling the null value returned by `getValidLinkGifts(pkm)`. Checking to see if the number is 6 is something the function does. If in the future the function was updated to support another generation, this would have to be changed as well, whereas nothing would have to change if you handled the null instead.

ReignOfComputer: Pokemon Link is a Gen 6 only feature, though. Hmm...

evandixon: In that case, that begs the question why a switch statement is used in getValidLinkGifts.

Oh, well. It works either way.

ReignOfComputer:
Moved it to return an empty EncounterLink instead of null.

But yeah, as a Gen 6 unique feature I suppose only one check is required?

PKHeX.Core/Legality/Encounters/EncounterGenerator.cs
kwsch: Was programmed before Gen7 came out, was expanded in anticipation that there would be a similar feature.

PR #1194, opened 2017-06-05
Past-Gen CurrentHandler Check

ReignOfComputer:
Check Save File Generation against PK6/7 Generation in case user's migrated Pokemon shares trainer details. Yes I'm guilty.

### DIFF
--- a/PKHeX.Core/PKM/PK6.cs
+++ b/PKHeX.Core/PKM/PK6.cs
@@ -480,19 +480,19 @@ namespace PKHeX.Core
         }
 
         // Synthetic Trading Logic
-        public void Trade(string SAV_Trainer, int SAV_TID, int SAV_SID, int SAV_COUNTRY, int SAV_REGION, int SAV_GENDER, int SAV_GENERATION, bool Bank, int Day = 1, int Month = 1, int Year = 2015)
+        public void Trade(string SAV_Trainer, int SAV_TID, int SAV_SID, int SAV_COUNTRY, int SAV_REGION, int SAV_GENDER, int SAV_VERSION, bool Bank, int Day = 1, int Month = 1, int Year = 2015)
         {
             // Eggs do not have any modifications done if they are traded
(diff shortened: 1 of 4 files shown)


kwsch:
Why not check version instead?

For the record, the games only check ID (4bytes), Gender, and OT via IsMyPokemon:

```
signed int __fastcall Savedata::MyStatus::IsMyPokemon(Savedata::MyStatus *this, const pml::pokepara::CoreParam *a2)
{
  Savedata::MyStatus *v2; // r4@1
  pml::pokepara::CoreParam *v3; // r6@1
  signed int v4; // r5@1
  bool v5; // r7@1
  const wchar_t *v6; // r2@3
  int v8; // [sp+0h] [bp-30h]@3

  v2 = this;
  v3 = a2;
  v4 = 0;
  v5 = pml::pokepara::CoreParam::StartFastMode(a2);
  if ( pml::pokepara::CoreParam::GetID(v3) == *(v2 + 2) && pml::pokepara::CoreParam::GetParentSex(v3) == *(v2 + 13) )
  {
    sub_49099C(v3, &v8, 13u);
    if ( gfl2::str::StrComp(&v8, v2 + 0x20, v6) )
      v4 = 1;
  }
  pml::pokepara::CoreParam::EndFastMode(v3, v5);
  return v4;
}
```

(if ID matches and gender matches, check if OT string matches), if so, CT=1 else 0.
It's feasibly impossible to match ID/Gender/OT across games without cheating... but we see it happen because people cheat :)

I'd rather remove the Country/Region checks as this OT/HT check was implemented before disassembly was done.

ReignOfComputer:
Made to check Version, sorry.

I'm guilty of cheating, my fun playthroughs use the same ID/SID for all my games >.<

kwsch:
Best to keep parity with the games and flag the TID/SID abusers :)

I double checked the current PKHeX code and it does the same check as the games (OT/ID/Gender), Country/Region is only supplied to facilitate geolocation history, so no changes necessary.

Issue #1200, opened 2017-06-07
"Evolution not valid" bug on move-based evolutions


White-Canary:
Move-based evolutions should be legal at the same level the move is learned, as the level-up event that triggers the learning of the move also triggers evolution with no further level-up required. Also, they should be legal at any level in gen 7 as the move reminder now teaches any level-up move regardless of current level.

![capture](https://user-images.githubusercontent.com/23516075/26899790-cd8d9afe-4b95-11e7-99d6-e01a124197f4.PNG)

javierhimura:
Ok, if i understand right that means
- Current level must be great or equal to the minimun learn level of the move, current check do not allow equal as legal.
- Minimun level for every move in generation 7 would be 2, that would be a pokemon learning the move at level 1 with move reminder and then level up to 2.

Still the current level must be one level after met level except for gen 3 non-eevee in generation 5 to 7.

PR #1201, opened 2017-06-07
Fix issue #1200 and added speculation for Gen2 VC without move reminder

javierhimura:
Two things
Fix issue #1200

Generation 2 VC without move reminder.
We can assume generation 2 wont have move reminder in the VC version because the only gen 2 game with move reminder is Pokemon Stadium 2, that means for generation 2 moves would work like in gen 1, only allowed the encounter moves and moves from one level after the minimun level to capture the pokemon

I have not done yet the part of move incompatibilities like generation 1. Is the part on gen 1 where we check a pokemon do not have two moves learned at the same level in redblue and yellos, also cheks that it do not have moves learned as an evolution at lower level that moves from a preevolutions. I talk about the functions ParseRedYellowIncompatibleMoves and ParseEvolutionsIncompatibleMoves. Also a third check should me made, a pokemon should not have a gen 1 move learned at a level greather than a gen 2 non-tradeback move.

I need to check how many moves combinations are illegal, like in generation 1 where there are not many and there is no problem to check one by one the illegal combinations.

### DIFF
--- a/PKHeX.Core/Legality/Core.cs
+++ b/PKHeX.Core/Legality/Core.cs
@@ -17,6 +17,7 @@ namespace PKHeX.Core
         public static bool AllowGen2VCTransfer => AllowGen1Tradeback;
         public static bool AllowGen2VCCrystal = false;
         public static bool AllowGen2Crystal => AllowGBCartEra || AllowGen2Crystal;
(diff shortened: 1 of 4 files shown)


kwsch:
The amount of arguments getting passed to the move validation methods is getting a little intense; may be worth refactoring into a passed object containing the move info.

PR #1205, opened 2017-06-09
Fix Pokémon GO, B2, and W2 origin game names

sora10pls:
Official name is Pokémon GO, not Pokémon Go; this is extremely minor. Met Location was already accounted for and doesn't need to be changed.

Edit: while I'm at it, I'll do B2/W2 as well.


sora10pls:
Let's pretend that second commit name says "B2" and not "BW"... :P

PR #1214, opened 2017-06-10
Do not allow backward transfer a pokemon without valid moves for destination game

javierhimura:
When a pokemon is converted from generation 2 to generation 1 any exclusive gen 2 moves are removed just like the real games do,  GSC does not allow a pokemon with gen 2 moves to enter the time capsule, but at the same time the games does not allow a pokemon to remove all the moves, it must have at least a non-empty move.

I just added the check for the case when a pokemon only have generation 2 moves, and its move list would be empy after removin all the moves, it should not be allowed to transfer because it wont have any move in generation 1

### DIFF
--- a/PKHeX.Core/PKM/PKMConverter.cs
+++ b/PKHeX.Core/PKM/PKMConverter.cs
@@ -221,7 +221,7 @@ namespace PKHeX.Core
                     case nameof(PK2):
                         if (PKMType == typeof (PK1))
                         {
-                            if (pk.Species > 151)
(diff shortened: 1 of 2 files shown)


kwsch:
Move compatibility is checked by `checkCompatible` a few lines down.

```
if (pk.Moves.Any(move => move > SAV.MaxMoveID))
{
    pk.Moves = pk.Moves.Select(move => move <= SAV.MaxMoveID ? move : 0).ToArray();
    pk.FixMoves();
}
```

kwsch:
Instead of transferring to 0 moves, can always insert the following after the `invalid == 0`:
```
if (invalid == 4)
    moves[0] = 1; // Pound
```

Issue #1161, opened 2017-05-22
Honey Tree modification corrupts save data


StealthyAqua:
On a Platinum save, any modification of the honey tree data causes the game to read the save data as corrupted upon loading. It then reverts to a version of the save that ignores all of the honey tree modifications.
![honey_tree](https://cloud.githubusercontent.com/assets/28866212/26311663/1d48a5f4-3ed3-11e7-857e-ccab0c012aad.png)

kwsch:
paging @suloku

kwsch:
Closed by #1215

PR #1215, opened 2017-06-10
fix Gen4 Misc Resources

pokecal:
add missing FlyDst and use MetLocation string.
add TrainerCard flags to FlagEditor.
add Pt resources.

fix HoneyTree, the BlockOffset needed.


kwsch:
Does this fix https://github.com/kwsch/PKHeX/issues/1161 ?

pokecal:
Fixed, if the corruption is caused by checksum error, and this checksum error is caused by offset error.

kwsch:
Thanks :)

Issue #1225, opened 2017-06-14
[Feature Request] Allow Item count modification for <= SAV3


sora10pls:
While giving all Items in <= SAV3 should be left out due to limited inventory slots, there should at least be a button to modify all Item counts.

This was added in ece845c but only applies to SAV4 forward due to the "Give All" button being removed for <= SAV3.

Edit: issued a pull request that resolves this. Would've gotten to it last night but desperately needed sleep. :P

PR #1226, opened 2017-06-14
Rework Inventory giveMenu

sora10pls:
- Allows for No Items/Item Count modification for <= SAV3 (hides Give All items function for <= SAV3)
- Replace "Give All" button text with "Modify" for clarity

Resolves my own issue #1225.

Example with an Emerald save:
image


sora10pls:
Will close this since you just committed your own changes, and there's no need to resolve conflicts now!

PR #1243, opened 2017-06-18
Update README.md

sora10pls:
PKHeX is capable of loading .sa1 and .sa2 file formats, which are used in some versions of the Visual Boy Advance emulator.

image


kwsch:
`sa*` is vba's way of having multiple save files for the same ROM; they're analagous with `sav`

http://ngemu.com/threads/is-there-a-vba-link-that-uses-sav-files-instead-of-sa1.77228/#post-1004069

sora10pls:
Thanks for clarifying :)

PR #1253, opened 2017-06-21
Fixed another potential blank error message

evandixon:
I also added comments where appropriate, because it took far too long to understand what was going on when reviewing the commit that fixed the original bug.

### DIFF
--- a/PKHeX.WinForms/Util/WinFormsUtil.cs
+++ b/PKHeX.WinForms/Util/WinFormsUtil.cs
@@ -228,8 +228,8 @@ namespace PKHeX.WinForms
             string pathCache = CyberGadgetUtil.GetCacheFolder();
             if (Directory.Exists(pathCache))
                 cgse = Path.Combine(pathCache);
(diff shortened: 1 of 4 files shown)


PKHeX.WinForms/Util/WinFormsUtil.cs
```diff
@@ -228,8 +228,8 @@ public static bool OpenSAVPKMDialog(string[] Extensions, out string path)
             string pathCache = CyberGadgetUtil.GetCacheFolder();
             if (Directory.Exists(pathCache))
                 cgse = Path.Combine(pathCache);
-            if (!PathUtilWindows.DetectSaveFile(out path, cgse))
-                Error(path);
+            if (!PathUtilWindows.DetectSaveFile(out path, cgse) && !string.IsNullOrEmpty(path))
```
evandixon: Looking more closely, path will never be null or empty when the function returns false. It may be a good idea to keep this here in case the function changes, but I can revert this part if you want. The important parts of this PR are really the comments.

Issue #1297, opened 2017-07-04
Various gen 4 issues


White-Canary:
- [x] Encounter type for Pokemon encountered in the Old Chateau should be "Building/Enigma Stone", not "None".
- [x] Setting the Pokedex to "Caught All" adds French language entries regardless of the actual game language.
- [x] Setting the Pokedex Upgrade to "National Dex" enables viewing the National Dex but doesn't enable any of the other things it unlocks (only observed in DPPt, not sure if it also applies to HGSS).
- [x] Unhatched Ranger Manaphy egg gives the error "Unable to match an event Egg encounter from the origin game".
- [x] Any Mail present in the bag in doesn't show up in PKHeX, and Mail added to the bag in PKHeX doesn't show up in-game (DPPt only, works fine in HGSS).
- [x] The various Mail items should use the correct in-game name instead of "Mail ## (G4)" (also applies to gen 5, but the gen 5 Mail items are different from the gen 4 ones despite having the same index numbers).

sora10pls:
Unable to replicate the Manaphy Egg bug.
![image](https://user-images.githubusercontent.com/17801814/27839749-bf749b8e-60c1-11e7-9985-9dc53f23c8fc.png)

As for Gen 4 mail, refer to source code's comment [on this line](https://github.com/kwsch/PKHeX/blob/5b5c9d00171681eb0576ddf1cd9810242a393d28/PKHeX.Core/Game/GameInfo.cs#L145).

sora10pls:
Should the Mail names be changed in PKHeX, here's a list of their in-game names:

Mail 1 -- Grass Mail
Mail 2 -- Flame Mail
Mail 3 -- Bubble Mail
Mail 4 -- Bloom Mail
Mail 5 -- Tunnel Mail
Mail 6 -- Steel Mail
Mail 7 -- Heart Mail
Mail 8 -- Snow Mail
Mail 9 -- Space Mail
Mail 10 - Air Mail
Mail 11 - Mosaic Mail
Mail 12 - Brick Mail

javierhimura:
To fix the old chateau encounter type the location should be removed from the list DPPt_MixInteriorExteriorLocations

kwsch:
Issue with mail is that it changed names from Gen4 -> Gen5

```
136	0x0088	Lustrous Orb
137	0x0089	Greet Mail
138	0x008A	Favored Mail
139	0x008B	RSVP Mail
140	0x008C	Thanks Mail
[...]

PR #1306, opened 2017-07-05
Actually fix Old Chateau Encounter Type

sora10pls:
Finally got around to actually testing this. The PKM in question is a legitimately caught Gastly in Platinum. My previous PR, as suggested by @javierhimura, was NOT the actual fix. This PR remedies this by checking for the correct Encounter Type, rather than not having one checked for at all.

Before **and** after first PR, PKHeX checked for "None", which was the bug initially brought up in #1297.
image

With this PR, it properly checks for "Building/Enigma Stone" and flags "None".
image
image

### DIFF
--- a/PKHeX.Core/Legality/Tables4.cs
+++ b/PKHeX.Core/Legality/Tables4.cs
@@ -228,7 +228,7 @@ namespace PKHeX.Core
         };
         internal static readonly int[] DPPt_MtCoronetExteriorEncounters =
         {
-            4, 5
+            4, 5, 70
         };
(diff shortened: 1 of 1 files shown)


javierhimura:
I dont understand when the change will be apply, DPPt_MtCoronetExteriorEncounters is only used when encounter type is assigned to mt coronet location, and there is not a 70 encounter table in mt coronet.

The buidling encounter type is assigned in function MarkDPPtEncounterTypeSlots

in this line

var GrassType = (Area.Location == 70) ? EncounterType.Building_EnigmaStone

PR #1307, opened 2017-07-05
Manaphy Egg fixes

sora10pls:
Resolves the 4th bug listed in #1297. Thanks @kwsch for providing the code!


kamronbatman:
@kwsch - I haven't checked yet, but what about the PID/IV relationship.  Do we want to include that in this code, or is that somewhere else & separate?

PR #1339, opened 2017-07-16
Added Location ComboBox

mohagan9:
Added into SimpleTrainer Dialog for Gen3.
Default = don't change current location.

I think this would be a good feature even across other generations, but with permission I would like to focus on attempting to implement this feature for Gen 3. 
The reasoning is that I am stuck on One Island and I would like to be able to get off without going through the required tasks! Also, it could give players the opportunity to explore areas earlier than allowed. 
 
image

### DIFF
--- a/PKHeX.WinForms/Subforms/Save Editors/SAV_SimpleTrainer.Designer.cs
+++ b/PKHeX.WinForms/Subforms/Save Editors/SAV_SimpleTrainer.Designer.cs
@@ -94,6 +94,8 @@
             this.LBL_BattleStyle = new System.Windows.Forms.Label();
             this.CB_BattleStyle = new System.Windows.Forms.ComboBox();
             this.CHK_BattleEffects = new System.Windows.Forms.CheckBox();
+            this.L_Location = new System.Windows.Forms.Label();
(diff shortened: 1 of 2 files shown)


kwsch:
Would have to be an MXY(Z) coordinate deal similar to Gen6+
image

Figuring out the locations of all the coordinates for every save file is a pain, hence why it hasn't been done by anyone yet.

note:escape rope, teleport, and fly are useful in-game features

mohagan9:
Well I was considering leaving out the XY co-ordinate aspect, and having a pre-established co-ordinates for each location to ensure the user didn't put in a co-ordinate that resulted in them being inside a wall for example.

For One Island you can't fly off it, and you can't fly to places you haven't visited and that don't have PC's so I do think it could be a useful feature.

Bl4ckSh4rk:
Would this even be possible at all? Gen 3 is storing tile map data of the location you last saved. So you would have to change that too. It's probably not a good idea to include all the maps in PKHeX.
Better use a warp AR code to get off the island.

mohagan9:
Ah okay, no worries. Thanks!

PR #1353, opened 2017-07-21
Fixed Zygarde Legalities from Gen 6 origin

### LINKED ISSUE #1351: Gen 6 native Zygarde form legality
Zygarde originating in gen 6 should be legal in 10% Forme with Power Construct, as the Zygarde Cube can change Power Construct Zygarde between its two forms.

architdate:
Fixes #1351

### DIFF
--- a/PKHeX.Core/Legality/Core.cs
+++ b/PKHeX.Core/Legality/Core.cs
@@ -2085,9 +2085,8 @@ namespace PKHeX.Core
                 return true;
             if (IsEvolvedFormChange(pkm))
                 return true;
-            if (pkm.Species == 718 && pkm.InhabitedGeneration(7))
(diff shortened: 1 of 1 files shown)


PKHeX.Core/Legality/Core.cs
```diff
@@ -2085,8 +2085,11 @@ internal static bool IsFormChangeable(PKM pkm, int species)
                 return true;
             if (IsEvolvedFormChange(pkm))
                 return true;
-            if (pkm.Species == 718 && pkm.InhabitedGeneration(7) && pkm.AltForm == 3)
```
kwsch: The interior check is basically `return pkm.AltForm != 1`.
It would be more succinct to replace `pkm.AltForm == 3` with `pkm.AltForm != 1`; same result but less lines.

architdate: Agreed and implemented

PR #1367, opened 2017-07-27
Fix Shaymin-Sky levelup learnset

### LINKED ISSUE #1355: Legality issues in 4th generation save
Shaymin, who learned air slash from a fourth-generation save file, is not legit.
Here is the pkm file.
This Shaymin is valid in 6th and 7th generation.
[492 - 쉐이미 - EA512C4A9A2B.zip](https://github.com/kwsch/PKHeX/files/1165323/492.-.-.EA512C4A9A2B.zip)

architdate:
Fixes #1355

### DIFF
--- a/PKHeX.Core/Legality/Core.cs
+++ b/PKHeX.Core/Legality/Core.cs
@@ -1767,6 +1767,10 @@ namespace PKHeX.Core
                             r.AddRange(LevelUpDP[index].GetMoves(lvl));
                             r.AddRange(LevelUpPt[index].GetMoves(lvl));
                             r.AddRange(LevelUpHGSS[index].GetMoves(lvl));
+                            if (species == 492) // Shaymin
+                            {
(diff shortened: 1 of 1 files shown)


javierhimura:
GetMoves already have a form atribute, it will be a better solution to use the form atribute in this line
int index = PersonalTable.HGSS.GetFormeIndex(species, 0);
Because not every form exits in all gen 4 games the  GetFormeIndex function should be called three times, for the index in HGSS, in Pt and DP, just like generation 5 have two form index for BW and B2W2, cheking for everi index if is not 0

javierhimura:
Just to clarify learnset do not needs to have a form atribute to have moves exclusive for one form, learnset is an array of moves learned by different species and forms, there are different elements in the array for every shaymin form 
PersonalTable.HGSS.GetFormeIndex(492, 0) and PersonalTable.HGSS.GetFormeIndex(492, 1) will return different form index.

Issue #1263, opened 2017-06-22
[Bug] Current Poketch App causes save data corruption


sora10pls:
- Editing current app corrupts save data and has the game load "the most recent save data"
- There is no "None" for current app; `0x00` is Digital Watch, not no app at all
![image](https://user-images.githubusercontent.com/17801814/27447973-02a4c336-5751-11e7-810b-541922b0b427.png)

Perhaps there's a checksum for this? I'd suspect `0x1160` but I could be wrong.

kwsch:
try checking pokesav to see how values are changed

sora10pls:
Before:
Offset `0x1162` == `0x06` (Dowsing Machine; App 7)
Offset `0xCF2A` == `0x6A`
Offset `0xCF2B` == `0x8E`

After:
Offset `0x1162` == `0x05` (Friendship Checker; App 6)
Offset `0xCF2A` == `0x61`
Offset `0xCF2B` == `0xC1`

sora10pls:
This may be useful reference: https://projectpokemon.org/wiki/Pok%C3%A9mon_NDS_Save_File_Checksum#Platinum_usage

kwsch:
https://github.com/kwsch/PKHeX/blob/427eb268e178a95e02c076c0afcc7876b51f4a01/PKHeX.Core/Saves/SAV4.cs#L74-L75

says it does a checksum 0x0000-0xCF18 and sets it to 0xCF2A

no idea why it didn't update on export...?

sora10pls:
After more testing, seems as if checksum is done for .sav, but not .dsv

sora10pls:
Okay, so I just did yet another test and when giving my save file all Poketch Apps, instead of updating the checksum at `0xCF2A`, it updated it at `0x4CF2A` and the save file loaded fine?

I assume that's the checksum for the backup save.

sora10pls:
OKAY forget everything mentioned before, turns out it was just some bug with the save file itself. After re-saving in game, the checksum began to properly get updated with PKHeX.

Only thing left to change is the fact that there is no "None" option for Current App.

sora10pls:
Yeah for some reason this keeps happening and then not happening, it's hard to fully replicate all the time.

kwsch:
Removed the "None" dropdown option so it now correlates directly; feel free to close if this ghostly issue isn't reproducable.

sora10pls:
Yeah I'm not too sure about this one, I'll go ahead and close.

PR #1416, opened 2017-08-25
Underground Score Editor (help needed)

egzonqj:
Hi! 

I've been looking for some Underground Score offsets through the Internet, I found them (atleast Pt ones, but I believe these are the same in DP) in [PP](https://projectpokemon.org/forums/forums/topic/14637-some-research-in-nds-save-files/) and wanted to put them in so everyone can edit their ones. 

I really think this could be written really better, because, for some reasons I don't get, when I'm trying to import the modified save, it says it's corrupted, yet by manually hex-editing the values and then fixing the checksums with PKHeX I managed to import the save properly (succeeded one time in my various attempts).

I'm opening the PR with all I've done, I hope I did not break anything else tho... sorry for my inexperience.

Here's a pic about what I've achieved
image

### DIFF
--- a/PKHeX.WinForms/Controls/SAV Editor/SAVEditor.cs
+++ b/PKHeX.WinForms/Controls/SAV Editor/SAVEditor.cs
@@ -1097,5 +1097,13 @@ namespace PKHeX.WinForms.Controls
             catch { }
             WinFormsUtil.Alert("Showdown Team (Battle Box) set to Clipboard.");
         }
+
+        private void B_OpenUGSEditor_Click(object sender, EventArgs e)
+        {
+            if (SAV.Generation!=4 && !SAV.DP || SAV.Generation != 4 && !SAV.Pt)
+                return;
+            else
+                new SAV_Underground(SAV).ShowDialog();
+        }
     }
 }
--- a/PKHeX.WinForms/Controls/SAV Editor/SAVEditor.cs
+++ b/PKHeX.WinForms/Controls/SAV Editor/SAVEditor.cs
@@ -1100,10 +1100,12 @@ namespace PKHeX.WinForms.Controls
 
         private void B_OpenUGSEditor_Click(object sender, EventArgs e)
         {
-            if (SAV.Generation!=4 && !SAV.DP || SAV.Generation != 4 && !SAV.Pt)
-                return;
-            else
-                new SAV_Underground(SAV).ShowDialog();
+            switch (SAV.Version)
+            {
+                case GameVersion.DP:
+                case GameVersion.Pt:
+                    new SAV_Underground(SAV).ShowDialog(); break;
+            }
         }
     }
 }
--- a/PKHeX.Core/Saves/SAV4.cs
+++ b/PKHeX.Core/Saves/SAV4.cs
@@ -242,7 +242,7 @@ namespace PKHeX.Core
                     Daycare = 0x1654 + GBO;
                     OFS_HONEY = 0x7F38 + GBO;
                     Box = 0xCF30 + SBO;
-
+                    
                     _currentPoketchApp = 0x1162;
                     break;
                 case GameVersion.HGSS:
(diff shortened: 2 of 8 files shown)


kwsch:
@sora10pls

sora10pls:
Opening this with my Platinum save throws an exception https://pastebin.com/wbY6wC2L

sora10pls:
Offsets listed in the PR are the exact same as PokeSAV's so it's working fine with Pt; gonna have to test D/P.

PKHeX.WinForms/Controls/SAV Editor/SAVEditor.cs
```diff
@@ -1097,5 +1097,13 @@ private void ClickShowdownExportBattleBox(object sender, EventArgs e)
             catch { }
             WinFormsUtil.Alert("Showdown Team (Battle Box) set to Clipboard.");
         }
+
+        private void B_OpenUGSEditor_Click(object sender, EventArgs e)
+        {
+            if (SAV.Generation!=4 && !SAV.DP || SAV.Generation != 4 && !SAV.Pt)
```
Tsunamical: if you wanna make it a bit cleaner
```
             switch (SAV.Version)
             {
                 case GameVersion.DP:
                 case GameVersion.Pt:
                     new SAV_Underground(SAV).ShowDialog(); break;
             }
```

sora10pls:
D/P offsets are different from Pt; source code works fine with Pt. D/P offsets below.

`0x3A2C`: Players Met Underground
`0x3A38`: Spheres Obtained
`0x3A3C`: Fossils Obtained
`0x3A30`: Gifts Given
`0x3A44`: Traps Hit
`0x3A48`: Traps Triggered
`0x3A60`: Flags Captured

Editing source code with the above offsets reveals that the offsets are valid for D/P (comparing PokeSAV).
image

sora10pls:
Hahaha @Tsunamical the save file I initially tested was all sorts of messed up, testing again with a good save file didn't throw an exception. Comment was deleted as a result.

Tsunamical:
Ahhhhh. Probably just want to use another switch then in SAV4.

sora10pls:
Refer to 33dff69

Tsunamical:
Ah now that's cool, it's been a bit since I looked at the setup.

sora10pls:
Should this get merged, I think the button should simply read "Underground" rather than "UG Scores". Minor nitpick

PKHeX.WinForms/Controls/SAV Editor/SAVEditor.cs
egzonqj: Oh, I'm going to do it this way, thanks!

egzonqj:
@sora10pls after editing, did you encounter the blue screen (corrupted save) when opening the game and importing the edited save? because that's what i was getting, gonna try now to see if it keeps happening..

sora10pls:
I am also getting the corrupted save file error. all I did was change my number of players met from 20 to 21. Upon file comparing in HxD, only that value was changed, and checksums were not fixed.

kwsch:
Possible culprit:
https://github.com/kwsch/PKHeX/blob/37b826d0ecbad2aaf9d4d5d07a9055d8905be74b/PKHeX.Core/Saves/SAV4.cs#L83-L91

Maybe it's not refreshing the checksums due to something in the save file not being expected?

sora10pls:
If you recall, this same issue happened with #1263; I initially thought it was just some fluke in me editing the save file, but what happened back then is happening again now.

egzonqj:
I guess there's something else to be done? I'll try to succeed in having a "Small block error" in the checksum check after editing so I can provide both saves

egzonqj:
Ok, I'm surprised I got it to work LOL, this is what I've done

1. Exported the save and edited manually three (0x3CB4, 0x3CC4, 0x3CE8) values with HxD
2. Opened the edited save with PKHeX, went to Verify Checksums and got "Small block checksum is invalid"
3. Saved the save and after doing it, went to Verify Checksums and got that they were valid
4. Imported the save and it prompted properly to the game

I'm comparing both (badchecksum and fixed) in HxD and this is the only thing that changes in block 0xCF2A-0xCF2B
image

Providing also the three saves (original, manually edited values -bad checksum-, pkhex checksum fixed)
[platinumtest.zip](https://github.com/kwsch/PKHeX/files/1252499/platinumtest.zip)

egzonqj:
I'm going to debug the fix procedure

egzonqj:
Ok, so when exporting the new save with the fixed checksum, the CRC16 checksum gave me 7191 (0x1C17) as a result and stored it at 0xCF2A.

Then I opened the new file and went straight to the checksum verification, and it gave the prior number, 7191, after that, I changed some values with the Underground Editor, and proceeded to export the save without anything else being changed. The CRC16 checksum gives 7733 (0x1E35).

After exporting the save, I compared both of them and 0xCF2A is edited with the new checksum result, with also the values edited with the UGEditor being there as well.

So I decided to import the save, and then it proceeded properly to the game, with the values ingame bein the new ones changed with the UGEditor.

After this, I tried to do it without debug, and everything worked fine by just editing some values. It worked perfectly! And tried one more time, this time editing some Pokémon in (and of course values of the UG Scores), and yep, it worked!

Not sure why it didn't update the checksum even if they were said to be valid after manually editing the values.

egzonqj:
I've tried both with a legit cartridge save and an emulated/r4 ROM one, and I think that the checksum isn't being copied in the save from the ROM.

The crc result I got while debugging that save (after editing a simple value with this underground editor) was 27418 (0x6B1A), the prior number stored in 0xCF2A was 59036 (0xE6CC), when exported the save and compared both files, only the value changed was there, while 0xCF2A kept being the same.

kwsch:
You're getting the random corruption due to not updating the correct save block; the GBO/SBO offsets are used to shift to the correct array location.

PR #1441, opened 2017-09-05
Revert PKHaX utility back to the old days

architdate:
Using executable name as a trigger for illegal mode would be more intuitive for normal users than writing and running a cmd/bash line (or creating a batch file)


sora10pls:
Why not both? :P

architdate:
You are not wrong to question this my friend @sora10pls

kwsch:
The executable shouldn't be renamed; you can always create a shortcut instead of a batch file, and name it PKHaX. When you choose to update the executable, there's no need to toggle between renaming, as the shortcut points to the new updated exe.

image

architdate:
Fair enough, i just thought having both options would be better for the enduser. But what you said makes perfect sense too. Thanks for clarifying this with me :)

PR #1456, opened 2017-09-15
Add JKSM backup path

sercho80:
For those using Rosalina and JKSM :)


sora10pls:
Does JKSM seriously export to that path with the 3DSX? I would've assumed it's still root -> JKSV -> Saves...

kwsch:
https://github.com/J-D-K/JKSM/blob/b6d67ac715ffac495cb212d27dc0d3ca8da5b4cf/source/sys.cpp#L102

architdate:
I don't know if this PR is being updated anymore, so I'll just send a PR to fix this myself @kwsch

PR #1467, opened 2017-09-19
Hide gender combobox in CXD

sora10pls:
Only male protagonists in Colosseum/XD (Wes and Michael respectively).


kwsch:
Gender still exists in the save file!
https://github.com/kwsch/PKHeX/blob/298dc344cfc388f31ae67206e7c65dff883a4b53/PKHeX.Core/Saves/SAV3Colosseum.cs#L356
https://github.com/kwsch/PKHeX/blob/298dc344cfc388f31ae67206e7c65dff883a4b53/PKHeX.Core/Saves/SAV3XD.cs#L241
No point hiding it ;)

sora10pls:
Strange how they kept it there for both games. For Pokemon it makes sense, Trainer I dunno :P

Issue #1443, opened 2017-09-06
OT Name too long for French pokémon from Gen 1


Pierrounator:
[French pokémon](https://gyazo.com/4d42cc839b59eccb576fee19e9306e82)

[The same with english language](https://gyazo.com/1e7141f0760e43cb24ff202bd7bda93a)

I'm playing Yellow and Blue right now and we can choose a nickname with 7 characters on both editions. So I think it's a bug from PKHeX. PKHeX allows only 5 characters in french.

There is the .PK1 file where I saw the bug

[001 - Bulbasaur.zip](https://github.com/kwsch/PKHeX/files/1281457/001.-.Bulbasaur.zip)

kwsch:
>New Update Available!

Anyway...

```c#
if (tr.Length > (lang == 2 ? 7 : 5))
```
How does transporter set the language?

Pierrounator:
I don't have the PokéBank so I don't know 😕

sora10pls:
Now's a good a time as any to buy a Pokemon Bank pass ;)

I'll test this in a bit.

sora10pls:
[Here](https://github.com/kwsch/PKHeX/files/1285820/fr_pika.zip) are two PKMs.

PK1 is the Pikachu from Yellow (French) before transferring.
PK7 is the Pikachu from Yellow (French) after transferring from Yellow to Moon.

All done via Poke Transporter and Pokemon Bank; no PKHeX whatsoever. @kwsch

PR #1486, opened 2017-09-23
Korean and Chinese restrictions

javierhimura:
Most of this checks are included because generation 1 and 2 games do not allow trade between games with different language, except games in western languages. With GB carts any trade between games with different RAM values result in save corruptions and VC games do not allow cross-language trade (not even between western languages)

Korean:
There is no Korean release for gen 1 pokemon, included VC games
https://bulbanews.bulbagarden.net/wiki/Pok%C3%A9mon_Gold,_Silver_coming_to_Nintendo_3DS_Virtual_Console
"The Generation I Virtual Console games were not released in South Korea either. " 
No gen1 games means any Korean gen2 is Gen2_NotTradeback, that means no gen1 origin nor moves are legal
Crystal was never released in Korean, probably when crystal is finally released in VC there will not be any Korean version. The checks for allowcrystal are changed to not allow crystal for Korean games from GB cart nor from VC with Crystal  VC allowed
Pokemon Stadium 2 was never released in Korean, also the games did not allow cross-language interaction, that means no move reminder for gen 2 Korean pokemon, even in GB cart saves.

Generation 4 can not trade between Korean and not Korean games, but Korean games can use the palpark with any language
https://bulbapedia.bulbagarden.net/wiki/Trade
"In Generation IV, Korean games cannot trade with non-Korean games due to only the Korean games including Hangul. Starting in Generation V, all games were made to be fully compatible with Korean games."

Chinese language restrictions
There is no Chinese release for gen 1 and 2 pokemon games, VC Chinese games are in Japanese
https://bulbanews.bulbagarden.net/wiki/Generation_I_Virtual_Console_games_released_in_Hong_Kong,_Taiwan
"Pokémon Red, Green, Blue, and Yellow are now available on Nintendo 3DS Virtual Console in Hong Kong and Taiwan. These games are in Japanese, as the original games were never released in Chinese. "

The source is from gen 1 games only but the games were never released in Korean nor Chinese. Nintendo has not made additional translations for VC games, it is very likely to assume china gen2 games are in Japanese too.

### DIFF
--- a/PKHeX.Core/Legality/Analysis.cs
+++ b/PKHeX.Core/Legality/Analysis.cs
@@ -219,7 +219,7 @@ namespace PKHeX.Core
             if (pkm.Format == 2 || pkm.VC2)
             {
                 // check for impossible tradeback scenarios
-                if (pkm.IsEgg || pkm.HasOriginalMetLocation || pkm.Species > Legal.MaxSpeciesID_1 && !Legal.FutureEvolutionsGen1.Contains(pkm.Species))
+                if (pkm.Korean || pkm.IsEgg || pkm.HasOriginalMetLocation || pkm.Species > Legal.MaxSpeciesID_1 && !Legal.FutureEvolutionsGen1.Contains(pkm.Species))
                     pkm.TradebackStatus = TradebackType.Gen2_NotTradeback;
                 else
                     pkm.TradebackStatus = TradebackType.Any;
--- a/PKHeX.Core/Legality/Analysis.cs
+++ b/PKHeX.Core/Legality/Analysis.cs
@@ -219,6 +219,7 @@ namespace PKHeX.Core
             if (pkm.Format == 2 || pkm.VC2)
             {
                 // check for impossible tradeback scenarios
+                // Korean gen2 games can't tradeback because there is no gen1 korean games released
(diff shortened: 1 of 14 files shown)


javierhimura:
I'm not sure if there is any legal restriction in western VC games. There is no way to trade between different languages but it is possible to nickname a pokemon to match another western language game and the GB data does not have IsNickname flag. The only possible exception will be in-game trade names, but we need first to have tables with the nicknames for GB games trade encounters.

kwsch:
iirc 1/2->7 doesn't detect the VC's language, it uses transporter's language setting. Would have to detect korean differently (using the `GetIsG2Korean` method) with the OT Name.

javierhimura:
I thought the gen7 language is based on the VC game used to trade to the bank, like for example english if the pokemon is transfer to gen7 from Pokemon Red and spanish if transfer from Pokemon Rojo.

But it that is not true i will change this tomorrow to use GetIsG2Korean method.

wwwwwwzx:
I use NA console and play VC Red in English. 1->7 will get `ENG` tag even if I change bank's and transporter's language to Chinese.
And I'm pretty sure there is no Chinese gen2 game.

sora10pls:
To add to this, back with #1443 the Pokemon transferred in that case had `FRE` tag when transferred with Transporter/Bank.

3DS, Transporter, and Bank are all Japanese (though Bank lets you choose language, and mine was set to English). Even with a French VC title, it still displayed as `FRE`.

PR #1510, opened 2017-10-04
Misc gen1/gen2 fixes

wwwwwwzx:
- G1/G2 only have Hiragana letter he -> fixed Houndoom(ヘルガー) /Heracross (ヘラクロス) default name mismatch
- Fixed Stadium2 Farfetch'd move (Thrash -> Slash)
- Added VC Mew TID check
- Update Chinese legality strings


kwsch:
Is the kata/hiragana issue fixable by changing the gen1/2 encoding table instead?
https://github.com/kwsch/PKHeX/blob/b5e3b616de0f8994e39e6be2c2202d2524afcbd7/PKHeX.Core/PKM/StringConverter.cs#L817

wwwwwwzx:
Yes. My change is just in case the user mixed up two characters in textboxs of nickname and OT.
```
0xCD => katakana ヘ
kata/hiragana ヘ => 0xCD
```
should also do the job.
The encoding tables are grouped in kata/hiragana, I think it's good to keep the order.

PR #1526, opened 2017-10-18
Update README screenshot

sora10pls:
Look mommy I'm an artist

(all I changed was the version string in the title)


kwsch:
Nothing visually changes besides the version string; shouldn't be a necessary update for every version (or every 6 months), but can accept for now.

Likely will be replaced in less than a month for US/UM content visuals.

PR #1541, opened 2017-10-23
Use manifest resources for string lists instead of the resources file

evandixon:
This allows external programs requesting string lists to get a string list, instead of an array with something like this:

```
..\Resources\text\en\text_Types_en.txt;System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089;Windows-1252
```

Tested by opening the form and by using this branch in the Project Pokémon Pokédex


kwsch:
Any noticeable difference in performance?

Calling the following every time a resource is requested seems kinda slow...
`typeof(Util).GetTypeInfo().Assembly.GetManifestResourceNames()` followed by O(n) search for string name
`typeof(Util).GetTypeInfo().Assembly`

Maybe building a dictionary when the library initializes (name string -> resource string) and fetching that way would be faster.

evandixon:
You're right. I ran Util.GetSpeciesList("en") 10K times, timing it with a stopwatch and recording the elapsed ticks, and it looks like the new code takes about 3 times what it take for the old code to work.

master:
* 23970193
* 24358695
* 22044843

f/string-resources:
* 64040648
* 63386879
* 64395218

By caching the assembly and manifest resource names, I managed to lower the execution time from 3x to 2x:
* 39139176
* 40887133
* 38068201

I then cached the result of each call to Util.GetStringList, and the execution time becomes 123196. Eyeballing the time it take for the main window to start between the two branches, these new changes feel faster, and hopefully this is not just me hoping it'll be faster.

I've also fixed a bug where an unlocalized language file wasn't falling back to English. Switching to certain languages no longer causes a crash.

PR #1599, opened 2017-11-13
add G7 MapUnmask & FlyDestination

pokecal:
for USUM. have not checked with SM, but I think it works.

### DIFF
--- a/PKHeX.WinForms/Subforms/Save Editors/Gen7/SAV_Trainer7.Designer.cs
+++ b/PKHeX.WinForms/Subforms/Save Editors/Gen7/SAV_Trainer7.Designer.cs
@@ -1,4 +1,4 @@
-namespace PKHeX.WinForms
+﻿namespace PKHeX.WinForms
 {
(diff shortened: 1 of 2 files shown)


pokecal:
I shortly update(push-f) for some fix.

kwsch:
Can you revise to use the event flag get/set similar to this recent commit? https://github.com/kwsch/PKHeX/commit/dbabcb9ef0a83f001478f137a221d60dc5e4d6a1

We've noticed the event flag values (less than 3968) are the same; any new event data uses any of the new 960 flags allocated. You'll probably get compatibility for both games by referencing the same flag ID, plus the lack of non-static offsets :)

pokecal:
OK

pokecal:
I think EventFlag can be separated 2 areas like this.
           : SM                                   : USUM
_event : start 0x23D0, 3072 flags : start 0x25D0, 4032 flags
_system : start 0x2550, 896 flags : start 0x27C8, 896 flags

PR #1610, opened 2017-11-15
Add Surfing Pikachu event flag

sora10pls:


kwsch:
It's tab separated (pls revise)

sora10pls:
I'll never understand why GitHub always breaks my tabs >_>

Messaged you on IRC as I pushed noting I realized the issue, but was late with the fix since I went out for a bit with a friend to lunch.

PR #1623, opened 2017-11-17
Flip Valid Trade OT Compare Operator

ReignOfComputer:
Related:
https://projectpokemon.org/home/forums/topic/42528-bw2-in-game-trades-yancycurtis-20171117/

### DIFF
--- a/PKHeX.Core/Legality/Checks.cs
+++ b/PKHeX.Core/Legality/Checks.cs
@@ -419,7 +419,7 @@ namespace PKHeX.Core
             if (pkm.IsNicknamed)
                 AddLine(Severity.Invalid, V9, CheckIdentifier.Nickname);
             int lang = pkm.Language;
-            if (validOT.Length >= lang)
+            if (validOT.Length <= lang)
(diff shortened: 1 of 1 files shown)


ReignOfComputer:
At least, I think this is correct >.<

PKHeX.Core/Legality/Checks.cs
```diff
@@ -419,7 +419,7 @@ private void VerifyTradeOTOnly(string[] validOT)
             if (pkm.IsNicknamed)
                 AddLine(Severity.Invalid, V9, CheckIdentifier.Nickname);
             int lang = pkm.Language;
-            if (validOT.Length >= lang)
```
evandixon: This looks like a guard for `validOT[lang]`. Wouldn't this cause issues if lang is actually greater than validOT.Length?

kwsch: it's more obvious if you flip the two arguments around
it's currently flagging if lang is less than, should be flagging if lang >= validOT.Length

kwsch:
test case pkm still fails as the OT is incorrect; need to verify the order of trainer langs for this

PR #1637, opened 2017-11-20
add ListBox to FestaPlazaEdit

pokecal:
for easy handling


pokecal:
shortly update with push -f

PR #1664, opened 2017-11-27
add Crystal GS EventFlags

pokecal:
have checked with US ver.


sora10pls:
Can double-confirm this works with US Crystal! Thank you :)

PR #1708, opened 2017-12-18
Create flags_xy_ja.txt

smileynation:
I can not understand the meaning of Statuette and can not translate.


sora10pls:
Statuettes mentioned in the flags are the mini statues received after reaching 50 Streaks in the Battle Maison.

smileynation:
@sora10pls Thanks.

Issue #505, opened 2016-11-26
Feature request: Pokeball and Shiny sprites


lomejor18:
it would be so great! so you can pick a matching ball... somebody is making those sprites:
https://github.com/msikma/pokesprite

![sin titulo](https://cloud.githubusercontent.com/assets/24194552/20642946/e0ac082c-b3f2-11e6-8feb-3b6ce504e5a5.png)

sora10pls:
Personally I too would like to see the Poke Ball images, but I feel like the shiny sprite addition isn't necessary. The star feels like enough to differentiate between shiny and non-shiny

lomejor18:
the star tells you that the pokemon is shiny, but you need to know the actual color to choose a matching ball... i'm an old player, but is difficult to remember every single shiny color

NinFanBoyFTW:
I would love to see this implemented. the pokeballs would be very useful especially for the obscure ones like friend ball or moon ball and I think it would be awesome to have the shiny sprites as long as it wouldnt take too much work since they already have most of them done from that link you posted.

PR #1735, opened 2018-01-02
Shiny Box Sprites

trainboy2019:


sora10pls:
Pretty sure @kwsch has commented on something like this in the past along the lines of "not official sprites, not gonna implement fan designs"

kwsch:
Not sure it's worth it when the shiny star already indicates. I'm usually against adding dependencies when they aren't necessary.
* Increases the exe size by ~800KB, or ~500KB when compressed. (+1/8th current size).
* These aren't official images
* Adds another project dependency (similar to QRCoder, same license)

Related discussion: https://github.com/kwsch/PKHeX/issues/505

Leaving open for now

ricksee:
If it could if it can't be added to the program, is there a way to add it ourselves? Similar to @architdate mod.

PR #1816, opened 2018-02-05
Fix build error

evandixon:
IEnumerable doesn't support LINQ methods, only the generic variant does. So this commit changes SAV_GameSelect's parameter to a generic IEnumerable of the only type being passed into it. I didn't realize it until afterward, but this is 100% the way it should have been because the databinding members expect ComboItems.


PKHeX.WinForms/MainWindow/Main.cs
```diff
@@ -674,7 +674,7 @@ private static GameVersion SelectMemoryCardSaveGame(SAV3GCMemoryCard MC)
             if (MC.HasRSBOX) games.Add(new ComboItem { Text = "RS Box", Value = (int)GameVersion.RSBOX });
 
             WinFormsUtil.Alert("Multiple games detected", "Select a game to edit.");
-            var dialog = new SAV_GameSelect(games.ToArray());
+            var dialog = new SAV_GameSelect(games);
```
evandixon: Not strictly necessary, but we're going to be calling .ToList anyway

PR #1840, opened 2018-02-17
Fix the build errors

architdate:
Pass SAV as a SAV4 into GetFormNames4Dex to fix no reference given to SAV_Pokedex4.SAV


architdate:
Wouldn't it be better to maintain the `SAV.Generation` consistency across the board? Regardless, thanks for the revert and fix :)

kwsch:
since the method is static and only used in gen4 cases (method is named after gen4)

the sav.generation ref was just lazy applying it to all areas that needed it (subforms)

PR #1857, opened 2018-03-02
Fix WC6/7 IV generation

wwwwwwzx:
IV lock doesn't work for IV3 case (Issue happens in Shiny Poipole WC7, all IVs besides HP are 31)
General WC  IVs check for [PGL Pikachu](https://github.com/projectpokemon/EventsGallery/blob/master/Released/Gen%206/Wondercards/JPN/1510%20ORAS%20-%20PGL%20Pikachu%20(JPN).wc6full) 
Source: ORAS: sub_45FB84 UMv1.2: sub_454408  https://pastebin.com/bbpS6svw

### DIFF
--- a/PKHeX.Core/Legality/Checks.cs
+++ b/PKHeX.Core/Legality/Checks.cs
@@ -590,15 +590,22 @@ namespace PKHeX.Core
                 if (IVs != null)
                 {
                     var pkIVs = pkm.IVs;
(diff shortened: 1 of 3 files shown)


kwsch:
I've noticed the same IV lock structure for static encounters; hopefully there aren't any that are incorrectly treated (ie forced to something other than 3IV)

PR #1866, opened 2018-03-19
Fix Suggested Move Set

ReignOfComputer:
Remove SelectedIndexChanged EventHandler to skip ValidateMove and allow moves to populate, then set back SelectedIndexChanged EventHandler

### DIFF
--- a/PKHeX.WinForms/Controls/PKM Editor/PKMEditor.cs
+++ b/PKHeX.WinForms/Controls/PKM Editor/PKMEditor.cs
@@ -669,15 +669,9 @@ namespace PKHeX.WinForms.Controls
 
             pkm.Moves = m;
             pkm.SetPPUps(m);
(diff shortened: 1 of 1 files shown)


kwsch:
Probably simpler to set FieldsLoaded around the `LoadMoves` call instead. When loading more than one pkm property, `FieldsLoaded` can be toggled on-off to disable the event handling update behavior.

PP Maxing isn't intended; I misused it (had wanted it to update the current PP counts with the new moves). Should probably split the functionality of the following method into two; one for PPUps 0/3 and one for updating the current PP:

https://github.com/kwsch/PKHeX/blob/d41df2420f7f7331c30e654e5232e88612fa1eba/PKHeX.Core/PKM/Editing/CommonEdits.cs#L151-L162

I can take care of this later today ;)

PR #1867, opened 2018-03-19
Fix PKMEditor HaX

ReignOfComputer:
Repeat detection of HaX in PKMEditor as the variable wasn't getting set anymore. Related: https://projectpokemon.org/home/forums/topic/44416-pkhax-not-working-in-the-latest-version/

### DIFF
--- a/PKHeX.WinForms/Controls/PKM Editor/PKMEditor.cs
+++ b/PKHeX.WinForms/Controls/PKM Editor/PKMEditor.cs
@@ -6,6 +6,8 @@ using System.Windows.Forms;
 using PKHeX.Core;
 using PKHeX.WinForms.Properties;
 using System.ComponentModel;
+using System.Diagnostics;
+using System.IO;
 
 namespace PKHeX.WinForms.Controls
 {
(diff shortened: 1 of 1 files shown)


kwsch:
The real problem is that HaX is passed from PKMEditor->StatEditor before the passing from Main->PKMEditor happens. 

I think the correct fix is to move this (in the form constructor):
https://github.com/kwsch/PKHeX/blob/102550174edaf6ecb8fd337649e3d5ba0311db0b/PKHeX.WinForms/Controls/PKM%20Editor/PKMEditor.cs#L44

To here:
https://github.com/kwsch/PKHeX/blob/102550174edaf6ecb8fd337649e3d5ba0311db0b/PKHeX.WinForms/Controls/PKM%20Editor/PKMEditor.cs#L127-L132

(not sure if this will cause any errors)

I also noticed (unrelated to this PR) that there's currently incorrect behavior toggling the Hacked Stats checkbox (never disables when unchecked due to it looking at the Enabled property instead of Checked):
https://github.com/kwsch/PKHeX/blob/cbf4038a950628bfc9ea5905ca4543da192aebd0/PKHeX.WinForms/Controls/PKM%20Editor/StatEditor.cs#L153-L159

ReignOfComputer:
Recreated in https://github.com/kwsch/PKHeX/pull/1869

PR #1871, opened 2018-03-20
Fix Suggested Moves PP for < 4 Moves

ReignOfComputer:
Uses FixMoves() to zero PP if move is blank.

### DIFF
--- a/PKHeX.WinForms/Controls/PKM Editor/PKMEditor.cs
+++ b/PKHeX.WinForms/Controls/PKM Editor/PKMEditor.cs
@@ -669,6 +669,7 @@ namespace PKHeX.WinForms.Controls
             }
 
             pkm.Moves = m;
+            pkm.FixMoves();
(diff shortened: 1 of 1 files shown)


ReignOfComputer:
Should have specified more clearly - not the PP, but the PP Ups.

kwsch:
pkm.SetMaximumPPCurrent() already sets the PP to zero if the move is 0... ?

ReignOfComputer:
Sets the PP to zero, but not the PP Up count.

Use my Floatzel (Ollie) with 4 moves (that is now a Beldum, so only Take Down is suggested as a move), and max PP. After applying suggested moves of only Take Down, the other 3 (None) moves still have max PP Ups.

[374 - Ollie - 48E3983106CE.zip](https://github.com/kwsch/PKHeX/files/1828134/374.-.Ollie.-.48E3983106CE.zip)

PR #1880, opened 2018-03-26
Detect event contest affection boost

### LINKED ISSUE #1879: Event Affection + Save Legality false flagging.
So there is currently false flags going up for event affection and I thought they should be addressed.

- If you have a gen 6 event and take it to a contest hall in ORAS the contest provides 20 affection to the original OT, not the current OT. What this means is that every contest you enter will trigger the pkhex legality flag because it currently expects all event affections to remain at the base.

The error on Pkhex reads "Invalid: Event OT Affection should be 0"

-If you have a gen 6 or 7 save that matches the event, such as if you manipulate the RNG for gen 6/7 and match the TID / SID / OT / Gender of the event for all intents and purposes you become the OT of that event. This currently will trigger a false legality flag in Pkhex. Furthermore if you ride around on a bike or feed it pokebeans in refresh it will also trigger more legality flags since it does not expect the base event happiness / affection to change.

For an event pokemon that matches the save it incorrectly flags "Invalid: Current Handler should not be Event OT"

If you rode around on a bike with the event pokemon matching your save and raised the friendship it will read incorrectly "Event OT friendship does not match base friendship"

If you fed it pokebeans in refresh / gave it cupcakes in amie it incorrectly reads "Event OT Affection should be 0"

If anything needs further clarification please let me know. Cheers.

sora10pls:
Resolves #1879

### DIFF
--- a/PKHeX.Core/Legality/Checks.cs
+++ b/PKHeX.Core/Legality/Checks.cs
@@ -1581,7 +1581,8 @@ namespace PKHeX.Core
             {
                 if (pkm.OT_Friendship != PersonalTable.AO[EncounterMatch.Species].BaseFriendship)
                     return new CheckResult(Severity.Invalid, V132, CheckIdentifier.History);
-                if (pkm.OT_Affection != 0)
+                // ORAS contests mistakenly apply 20 affection to the OT instead of the current handler's value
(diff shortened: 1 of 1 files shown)


kwsch:
Would prefer if the !=0 check is performed first (for performance reasons)

sora10pls:
done :)

PR #1885, opened 2018-04-03
Update Chinese translation file

wwwwwwzx:


wwwwwwzx:
Something I noticed:
- One redundant line: 
https://github.com/kwsch/PKHeX/blob/ad19496d9437e500436792cc7ce1039db9f2cf33/PKHeX.Core/Resources/text/en/lang_en.txt#L213
- Some of the buttons in FLP_SAVtools are missing entry, like Main.B_CellsStickers
- Are the flag translations of ORAS and XY still necessary?

kwsch:
The brute forcing to retrieve all translatable controls dynamically creates all forms with a minimum of one (1) constructor argument; the Main form (which contains the SAV & PKM editors) has no parameters -> skipped.

I was triggering the brute force method from the main form, which ended up creating another main form, which triggered the brute force again... etc (yay recursion). Will probably just relocate dev functions to a new form.

Will need to finish the removal method (basically does the translation but removes any entries that are translated, any that remain at the end are unused -> delete from txt). That should resolve the unused flag labels & redundant lines.

PR #1887, opened 2018-04-05
Translations update

wwwwwwzx:
- Update Chinese translations
- Fix base stats label name
- Add back translations lost by case mismatch


kwsch:
heh that definitely clears up my confusion on "Base" 😆

kwsch:
I'll probably start another refactor to rip out message strings used by the GUI (thus the program will be 100% translatable) whenever I get bored with usual updates.

Might end up the same as legality translations (static class -> reflection from strings to set)

PR #1892, opened 2018-04-08
Translated to spanish some of the lines (from 1 to 160) of MessageStrings

egzonqj:


egzonqj:
I would like to add that I'm not a native spanish speaker, so some things could be wrongly translated, yet I put all my efforts to do it as good as it should be.

Issue #1856, opened 2018-03-01
Request: autodetect *.SaveRAM extension


Ammako:
Much like how .dat and .dsv are detected.
This extension is used by BizHawk. Relevant for Gen. 1, 2 and 3.

One *could* just manually delete the extra ".sav" in the save dialog every time before saving, or remove it afterwards in file explorer, but for convenience reasons it'd be nice if the extension could just be detected as such and avoid adding an extra ".sav" at the end of the file name.

(Worth mentioning that, unlike DeSmuME, BizHawk uses raw save files, it just uses a different file extension. So the file contents are the same as you'd find in .sav or .dat, no header or footer or anything.)

kwsch:
PKHeX will not support loading/editing of save states:
* RAM data is not laid out in the same manner as the stored savedata.
* Every emulator can have a different internal format.
* Internal formats can change between emulator versions.

Ammako:
It's not save states lol.

Ammako:
I didn't think I'd need to since I already specified that it's the exact same as .sav just in a different extension, much like how VC .dat is the same file contents as .sav, but

[Pokemon - Red Version (USA, Europe).zip](https://github.com/kwsch/PKHeX/files/1769489/Pokemon.-.Red.Version.USA.Europe.zip)

kwsch:
File loads just fine via drag/drop, or opening with the file->open menu.
PKHeX doesn't care about file extensions; the only exception to this is checking ".dat" for determining if the save originated from a VC game.

Not going to add filters for every extension for every emulator that has their own unique file naming conventions. The All Files option is the default for file->open and allows for anything to be opened.

Ammako:
>Not going to add filters for every extension for every emulator that has their own unique file naming conventions.

So I guess DeSmuME just gets special treatment because reasons?

kwsch:
It doesn't get special treatment:
![image](https://user-images.githubusercontent.com/6393368/36823438-065707ec-1cb2-11e8-889d-0245b2ffcbe2.png)

Ammako:
Demonstrably wrong.

![image](https://puu.sh/zxWLL.png)

Ammako:
(Worth noting that .dat save files get "Save as type: SAV File (*.sav)" on the save as dialog, despite keeping their .dat extension as-is without getting an extra .sav added onto the end. However it's being done for .dat, it doesn't have to require adding a whole new file type extension like for DeSmuME DSV.)

Issue #1455, opened 2017-09-14
Support SRM RetroArch save files


megadrifter-pt:
**RetroArch** uses a common save file format for all of it cores, with the extension _srm_ (Save RAM), including **Game Boy Color** cores (TGB Dual, Gambatte, etc).

Some of us may use this frontend to play on emulators. Would you add some support for those files ?

kwsch:
PKHeX won't support save states, as RAM snapshots are not save files. Just retrieve the savegame after saving ingame.

megadrifter-pt:
There are two kinds of SRM files. Those which are save states (found in _states_ folder), and those which are saves generated by games (found in _saves_ folder). The latter looks a lot like SAV files (if not exactly the same). Changing the extension to _sav_ does the job to open it with PKHeX.
___
Thanks a lot for this amazing tool!

kwsch:
PKHeX doesn't care about extensions, you can always feed it those files directly (selecting all files filter or just dropping it in).

Issue #1130, opened 2017-05-14
Possible Option: Allow Saving Files As Other Formats 


ricksee:
Is it possible that a conversion between supported formats is possible. Like changing .sav to .dsv for instance when exporting or is it a step simple as changing the file extension name.

sora10pls:
What would the benefit of this be? DeSmuME allows you to import raw save files (.sav) directly through the "File" dropdown. Upon doing so, it immediately adds a .dsv equivalent in the "Battery" (saves) directory.

ricksee:
It may be easier to do it on one program than relying on desume. Plus, .dsv wouldnt work for me when doing this.

sora10pls:
As I said, DeSmuME imports .sav and immediately converts to .dsv

If you want to import a .dsv, just replace the file in your "Battery" directory => <same name as ROM>.dsv

Decision is always up to @kwsch and I understand its convenience, but I personally just find no use for it.

PR #1955, opened 2018-05-21
Add *.fla files to file dialogs

MichiS97:
This adds .fla save files created by the Everdrive GBA flashcard to the list of supported files to make handling these files a little more convenient.

### DIFF
--- a/PKHeX.Core/Saves/SAV3.cs
+++ b/PKHeX.Core/Saves/SAV3.cs
@@ -10,7 +10,7 @@ namespace PKHeX.Core
     public sealed class SAV3 : SaveFile
     {
         public override string BAKName => $"{FileName} [{OT} ({Version}) - {PlayTimeString}].bak";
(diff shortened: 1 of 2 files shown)


kwsch:
Not sure if this is the best way to go about it; there's so many emulators/etc with their own file extensions, it'd be a mess to start including every single one. 

The only exemptions are 

- .sav (universal "SAVefile) extension)
- .dat (official VC extension)
- .gci (GameCube Image; dolphin is the only gc emu afaik and is consistent).

https://github.com/kwsch/PKHeX/issues/1856
https://github.com/kwsch/PKHeX/issues/1455
https://github.com/kwsch/PKHeX/issues/1130

Might be better to update the export filter handling to detect if the extension is different from standard, and update the import filter to load in custom user filters from a text file?

MichiS97:
Hm yes, I see your point.

PR #1957, opened 2018-05-22
Correct Control Switching for G2->G7

ReignOfComputer:
Is there a reason for having GSC newTrack as USUM? This causes `origintrack = newTrack;` not to happen and thus controls don't switch from G2->G7 correctly.


kwsch:
G2 origin pkm should show gen7 met locations

kwsch:
I can't see anything wrong in the behavior if I removed those lines, so okay 👍

PR #1958, opened 2018-05-22
Stop Hyper Training Manual IVs

ReignOfComputer:
Importing a set with deliberately low IVs (e.g. "IVs: 0 Atk") still undergoes automatic Hyper Training. This commit stops application of Hyper Training if IVs are specified. If Hidden Power is declared, it checks if there are any IVs below 30 to decide on Hyper Training. Should cover /most/ cases.

### DIFF
--- a/PKHeX.Core/Editing/CommonEdits.cs
+++ b/PKHeX.Core/Editing/CommonEdits.cs
@@ -311,7 +311,10 @@ namespace PKHeX.Core
             pk.IVs = Set.IVs;
             pk.EVs = Set.EVs;
 
-            pk.SetSuggestedHyperTrainingData(Set.IVs);
+            pk.HyperTrainFlags = 0;
+
+            if (!Set.ManualIVs)
+                pk.SetSuggestedHyperTrainingData(Set.IVs);
             if (ShowdownSetIVMarkings)
                 pk.SetMarkings();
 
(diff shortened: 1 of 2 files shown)


kwsch:
`SetSuggestedHyperTrainingData` already sets the Hyper Training flags contextually based on the final IVs:

https://github.com/kwsch/PKHeX/blob/97e21ce133ea2ab99a3edf628fc25dae3c6197c4/PKHeX.Core/Editing/CommonEdits.cs#L233-L256

I'm able to import sets with 0 IVs and they don't get hyper trained. Speed and Attack are the only exclusions (checking > 2).

```
Amoonguss (M)
IVs: 0 Atk / 1 Def / 0 SpD / 1 Spe
Ability: Effect Spore
Calm Nature
- Energy Ball
- Giga Drain
- Hidden Power [Ghost]
- Protect
```
image

ReignOfComputer:
Doing something like

Ollie (Floatzel) (M) @ Mystic Water
IVs: 0 HP / 0 Atk / 0 Def / 0 SpA / 0 SpD / 0 Spe
EVs: 6 HP / 252 Atk / 252 Spe
Ability: Swift Swim
Jolly Nature
- Ice Punch
- Return
- Waterfall
- Crunch

Ends up with HP, Def, SpA, SpD all Hyper Trained.

kwsch:
If the user defines a hidden power type with 0 attack IVs & other '30' IVs, the proposed logic would not hyper train the 30s->31.

Sure there's rare cases where you could want a non-flawless stat (Seismic Toss number, Endeavor) but it'd be much less frequent than Hidden Power with 0atk/spe. Gotta let the user opt out manually in that scenario ;)

PR #2061, opened 2018-07-16
Update README.md

kang1806:
I am new to the Git Community and a beginner in programming, i just hope that you will accept this contribution, Thx


kwsch:
Bullet points are already separate enough; adding 3 commas isn't really value added.

PR #2117, opened 2018-09-12
Add XY Magikarp trade IVs

### LINKED ISSUE #2116: XY In-game Trade Magikarp
This Magikarp has static 31 IVs for its Attack and Special Attack stats. The legality checker does not account for this and allows it to pass, it seems.

Here's a screencap:
untitled-1

If you need proof, I'm providing.

Here are RoC's:
https://github.com/ReignOfComputer/RoCs-PC/blob/master/19%20-%20Gen%20VI%20-%20XY%20Collection/Pokemon%20X/In-game%20Trades/129%20-%20Magikarp%20%7BCarpe%20Diem%7D%20-%209976D0381382.pk6

https://github.com/ReignOfComputer/RoCs-PC/blob/master/19%20-%20Gen%20VI%20-%20XY%20Collection/Pokemon%20Y/In-game%20Trades/129%20-%20Magikarp%20%7BCarpe%20Diem%7D%20-%202F4D2C492993.pk6

And finally, mine:
[129 - Carpe Diem - 5C5B8AF53A2C.zip](https://github.com/kwsch/PKHeX/files/2373661/129.-.Carpe.Diem.-.5C5B8AF53A2C.zip)

sora10pls:
Resolves #2116.

### DIFF
--- a/PKHeX.Core/Legality/Encounters/Data/Encounters6.cs
+++ b/PKHeX.Core/Legality/Encounters/Data/Encounters6.cs
@@ -379,7 +379,7 @@ namespace PKHeX.Core
         #region Trade Tables
         internal static readonly EncounterTrade[] TradeGift_XY =
         {
-            new EncounterTrade { Species = 129, Level = 5, Ability = 1, Gender = 0, TID = 44285, Nature = Nature.Adamant, }, // Magikarp
+            new EncounterTrade { Species = 129, Level = 5, Ability = 1, Gender = 0, TID = 44285, Nature = Nature.Adamant, IVs = new[] {-1, 31, -1, -1, 31, -1 }, }, // Magikarp
             new EncounterTrade { Species = 133, Level = 5, Ability = 1, Gender = 1, TID = 29294, Nature = Nature.Docile, }, // Eevee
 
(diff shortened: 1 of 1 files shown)


kwsch:
Are there other fixed IV cases that are missing? Verify with pk3DS

sora10pls:
Gifts are accounted for already, though there are two things of note:

- IV3 applies to Cosplay Pikachu, but Speed is guaranteed one of those 3.
https://github.com/kwsch/PKHeX/blob/92489ad2062ea2e7c165d7b3a7c57659fbfd7285/PKHeX.Core/Legality/Encounters/Data/Encounters6.cs#L364-L368
image

- Trades can't be viewed/modified with pk3DS for Gen 6 (we've talked about this in the past, just haven't found them yet), but according to multiple sources their IVs are accounted for in `Encounters6.cs` (XY Eevee has no guaranteed perfect IVs).

kwsch:
Maybe adding the `IVs = new[] {-1,-1,-1,31,-1,-1}` for the Cosplay Pikachu would be okay.

https://github.com/kwsch/PKHeX/blob/789449f4d6512c49fdc3105446d0590663d1507c/PKHeX.Core/Legality/Encounters/Generator/EncounterStaticGenerator.cs#L181-L188

IV3 is checked after the encounter passes initial validation; individual IV requirements are checked before the encounter is validated. In other words, they won't override eachother as they're checked at different stages.

PR #2131, opened 2018-09-30
Allow viewing and editing of Generation II held items from a Generation I save

soopercool101:


soopercool101:
Forgot a few extra edits. Moved to #2132

PR #2132, opened 2018-09-30
Allow editing of Generation II held items from a Generation I save file

soopercool101:
Among other things, allows fixing legality of PK1 Pokemon with Generation II tradeback moves without exporting to a PK2 and reimporting.


soopercool101:
Perhaps adding a numeric input for editing the Catch_Rate value would be a more elegant solution in that case?

PR #2150, opened 2018-10-31
Update README.md

sora10pls:
Checkpoint changed its repo ownership. The forum post should also be changed to fully match the GitHub's readme, as some bits are inconsistent (still reference svdt instead of Checkpoint).


kwsch:
It technically already redirects, but whatever ;)

PR #2151, opened 2018-11-01
Fix Electabuzz data

Admiral-Fish:


PKHeX.Core/Legality/Encounters/Data/Encounters3Shadow.cs
```diff
@@ -678,8 +678,8 @@ public static class Encounters3Shadow
             Species = 125, // Electabuzz
             Locks = new[] {
                 new NPCLock(277), // Shadow Swellow
-                new NPCLock(065, 12, 0, 063), // Alakazam (M) (Serious)
-                new NPCLock(230, 18, 1, 127), // Kingdra (F) (Bashful)
+                new NPCLock(065, 24, 0, 063), // Alakazam (M) (Serious)
```
kwsch: Comments didn't get updated; can be updated later.

PR #2155, opened 2018-11-02
Fixes a unit test that gets Pokémon names

evandixon:
There were additional Pokémon names added, but the test was never updated :(

After some infrastructure changes, I re-enabled automatically building pull requests, which should give more visibility on this in the future (hopefully).


evandixon:
Maybe now would be a good time to move the tests to .Net Core

evandixon:
I'll reopen this PR after I'm finished

PR #2156, opened 2018-11-02
Port tests to .Net Core

evandixon:
This removes the need to use Windows to run tests

Caveat: one test is failing, and I'm not sure what I broke, or if something was already broken. It works in master, but I didn't see any important changes outside the tests


Tests/PKHeX.Core.Tests/Simulator/ShowdownSetTests.cs
```diff
@@ -184,7 +175,8 @@ public void TestGenerate()
                 foreach (var e in ez)
                 {
                     var la = new LegalityAnalysis(e);
-                    Assert.IsTrue(la.Valid);
+                    la.Valid.Should().BeTrue($"Because all generated Pokemon for {i:000} should be valid");
```
evandixon: This is the part of the test is failing, where i=12

Tests/PKHeX.Core.Tests/Simulator/ShowdownSetTests.cs
```diff
@@ -17,160 +17,151 @@ static ShowdownSetTests()
...
+            Assert.True(!sets.Any());
 
             sets = ShowdownSet.GetShowdownSets(new [] {"", "   ", " "});
-            Assert.IsTrue(!sets.Any());
+            Assert.True(!sets.Any());
         }
 
-        //[TestMethod]
-        //[TestCategory(SimulatorParse)]
+        [Fact]
```
evandixon: I could make this a theory, so we have one test per loop iteration. This way, we know how many things fail when there's at least one failure.

Tests/PKHeX.Core.Tests/Simulator/ShowdownSetTests.cs
kwsch: I had left it as a debug way test IEncounterable -> pkm code back a while ago.

I had commented out (see `//[TestCategory(SimulatorParse)]`) thus the test was never part of the usual test suite. With the recent addition of shadow lock checks, it's likely failing because of that.

evandixon: > I had commented out (see //[TestCategory(SimulatorParse)]) 

Ah, right, I think I remember seeing something like that.

> Is it possible to just skip this method for testing like before?

Yep. I can either use the Skip attribute or just comment it out. I'll decide which to use when I push an update

PKHeX.Core/Util/DataUtil.cs
```diff
@@ -85,7 +87,12 @@ public static string[] GetStringList(string f)
             string[] rawlist = txt.Split('\n');
             for (int i = 0; i < rawlist.Length; i++)
                 rawlist[i] = rawlist[i].TrimEnd('\r');
-            stringListCache.Add(f, rawlist);
+
+            lock (getStringListLoadLock) // Seems to fix an issue where concurrently-running tests get a null from this function
```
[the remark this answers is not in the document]
evandixon: I took another look at it, and after adding an additional safety check, I think it'll be fine. It's certainly conceivable two threads could hit that locks at once

evandixon:
I approve of my own PR pending status checks. If they all pass, I'll merge it in a day or two to give people time to review.

PR #2163, opened 2018-11-14
Enhancement - Double click on Pokemon to view

Yushell:
Added some code to enable viewing of Pokemon when double clicking on the sprite inside the BoxEditor. It helps speed things up when editing Pokemon.


kwsch:
Seems like it's duplicate with the already existing click-hotkeys:
https://github.com/kwsch/PKHeX/blob/18f5ef7c7b134bb1872061e7a82fa38d171fed5a/PKHeX.WinForms/Resources/text/shortcuts.txt#L63-L65

Yushell:
My bad. Did not notice those shortcuts existed :)

Issue #2158, opened 2018-11-08
Put PKHeX.Core on NuGet


leo60228:
This would make it easy to include in applications. Assuming the API is unstable (which I'm not sure it is), it would also easily allow pinning to a specific API version via Semver. This especially makes sense now that it's tested on .NET Core, which I don't believe PKHeX uses.

kwsch:
The winforms exe uses version.txt to:
* Display the release date in the program title
* Compares to github address to check for updates. 

If there's an alternative for both of those that doesn't introduce dependencies, then I'm okay with adjusting things.

evandixon:
While I work on finding a solution, PKHeX can be found on this MyGet feed:
https://www.myget.org/F/projectpokemon/api/v3/index.json

The version numbers don't really mean anything beyond what TC build it's from. I'll work on fixing that soonish.

leo60228:
Thanks!

kwsch:
Done!
https://www.nuget.org/packages/PKHeX.Core/

PR #2164, opened 2018-11-14
Use assembly version as the current version

evandixon:
I updated the Team City configurations to use GitVersion, and now the resulting assemblies' versions will match whatever the latest tag is + 1 or more patch versions based on the number of commits and exact config.

The intended flow for releases is now:
1. Tag the branch with whatever Major.Minor.Patch version you want. Feel free to make it based on the date or not.
2. Upload the build artifacts from TeamCity to the site, so people can download it. If you choose to build it yourself, you'll need to update the assembly version in AssemblyInfo.cs.

Point 2 can be replaced in the future with [one of my projects](https://github.com/evandixon/IPS-DotNet-Client), though it could use a little more work to support things like generating changelogs.

This PR is part of #2158, since having a robust versioning system is important when creating NuGet packages.


PKHeX.WinForms/MainWindow/Main.cs
```diff
@@ -103,7 +106,7 @@ private set
         private static string TemplatePath => Path.Combine(WorkingDirectory, "template");
         private static string PluginPath => Path.Combine(WorkingDirectory, "plugins");
         private const string ThreadPath = "https://projectpokemon.org/pkhex/";
-        private const string VersionPath = "https://raw.githubusercontent.com/kwsch/PKHeX/master/PKHeX.WinForms/Resources/text/version.txt";
```
evandixon: Can't get rid of the old version.txt until people have mostly updated to the new version

PKHeX.WinForms/MainWindow/Main.cs
```diff
@@ -103,7 +106,7 @@ private set
...
+        private const string VersionPath = "https://raw.githubusercontent.com/kwsch/PKHeX/master/version.txt";
```
evandixon: Since the old file can't go away, maybe I should just go ahead and move the new version to a subdomain of projectpokemon.org or something.

version.txt
```diff
@@ -0,0 +1 @@
+18.11.01
```
evandixon: This corresponds to a tag I made on the commit from the last release. If the premise of the PR is abandoned, feel free to delete the tag.

evandixon:
Updated to use the GitHub API. I feel a lot better about this now.

PR #2205, opened 2018-12-07
French Translation update

Ntonio36:
Here I come again 😄


kwsch:
Did you double check the translated UI that it doesn't overflow the space it is allowed?

https://github.com/kwsch/PKHeX/pull/2205/files#diff-751bfb54eab4a4490e6b6fdad904d8e1R1130

PR #2248, opened 2019-01-25
add offsets & save-size for ranch's platinum update

PMArkive:
I don't know yet if the OT offset is the same though


PMArkive:
I have two save files from Dolphin which are rather bare since Dolphin does not support DS connection.
[ranchtests.zip](https://github.com/kwsch/PKHeX/files/2798293/ranchtests.zip)

PR #2250, opened 2019-01-26
Add USUM Wormhole functions

PMArkive:
The current wormhole's shiny flag & pokemon can be changed with these functions.

The flags & offsets were found by @PP-theSLAYER https://projectpokemon.org/home/forums/topic/39433-gen-7-save-research-thread/?page=3&tab=comments#comment-239090

I haven't comprehensively tested each slot and wormhole to confirm the correctness (yet). This is assuming theSLAYER's table was correct.

I didn't make a subform because I've never made GUIs in C# and because I want to see if this code needs to be changed first (like the slot base, 0, or max variable, 4). Maybe more comments too...


kwsch:
Might be better maintainable to have it as a "WormholeInfoReader" of sorts, similar to how LGPE was done with separate block readers. I can refactor, but the GUI portion is also ehh

PR #2271, opened 2019-03-15
Add PlayerNames and MatchStamp to BV7

bsinky:
The parsing logic was copied from BV6, and just the offsets updated for Gen 7. I've only tested this against Battle Video files generated by Ultra Moon, so I'm honestly not sure if the non-Ultra games Battle Videos are any different.

Let me know if you'd like me to do any refactoring of the copied code from BV6 so it can be shared between the two classes, or if I should pursue additional testing, thanks!

### DIFF
--- a/PKHeX.Core/Saves/Substructures/Battle Videos/BV7.cs
+++ b/PKHeX.Core/Saves/Substructures/Battle Videos/BV7.cs
@@ -1,4 +1,6 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
+using System.Text;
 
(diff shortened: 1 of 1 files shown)


kwsch:
The battle video format is the same for a given generation (eg SM<->USUM, likewise XY<->ORAS), so there won't be any differences.

In terms of establishing a base class for gen6/7, it probably won't be worth the effort to just reduce code duplication. For PKHeX.Core usage, extra properties are fun to have, but the PKHeX.WinForms (GUI) just cares about ripping PKM data out. That said, feel free to expand things if you want!

Issue #2298, opened 2019-05-11
Shiny Ash-Greninja false positive


DaWoblefet:
As far as I know, Ash-Greninja can't be shiny; I was sent this pk7 of a Shiny Ash-Greninja that passes PKHeX's legality checker on the latest release.
![image](https://user-images.githubusercontent.com/23667022/57573024-ebaaf500-73e7-11e9-934c-058500d04266.png)
[658-01_-_Greninja_-_E95300000021.zip](https://github.com/kwsch/PKHeX/files/3169309/658-01_-_Greninja_-_E95300000021.zip)

AbdallahKATAIE:
I'm a bit concerned about this.
Is it possible that there are some other forgotten cases ? if yes, how can I make sure that all my Pokémon are legal even though PkHex says they are ?

WardenUnleashed:
Ash-Geninja's shiny check is a bit trickier than most as Ash-Greninja is not Shiny locked as an event however its PID is overwritten upon transfer to from the demo to SM.

I've written a bug fix for this issue, that will check to ensure that Ash-Geninja is not shiny.

PR #2299, opened 2019-05-11
Issue 2298: Add shiny check for Ash-Greninja to Legality Validator

WardenUnleashed:
https://github.com/kwsch/PKHeX/issues/2298

It was noted in this issue that Ash-Greninja was passing legality checks when shiny. While the event card for Ash-Greninja is not shiny locked, when Ash-Greninja is transferred from the demo to S/M, its PID is overwritten and forced to be non-shiny, effectively shiny locking it.

The added logic reflects this constraint


WardenUnleashed:
Pretty sure those build issues are unrelated to my changes.

PR #2305, opened 2019-05-13
Enhance Invalid Convert Error

ReignOfComputer:
Makes it clearer for conversion failure with Pokemon files - prevent confusion with save files.


kwsch:
Thanks!
I'll eventually switch the hardcoded array to use ExtraSaveExtensions instead.

PR #2319, opened 2019-06-09
Split gen5-7 saves with inheritance

kwsch:


kwsch:
* `new SaveFile*()` and `new SaveFile*(data)` are now distinct constructors; lack of data will result in a save that can't be exported, like before. Providing data will allow export. This skips checking for null & checking for data presence.
* Block offsets are now hardcoded for gen5+ and shared (static). Individual block chunks now perform the majority of the logic, in order to simplify the main savefile object. Example: PokeDex logic is now done in the blocks.
* WC6 eon ticket value now is applied via the Mystery Gift block inserting the ticket. It now also enables it to be retransmitted (new).
* Gen6-7 saves now inherit from SAV_BEEF, which shares the beef block (checksum) implementation.

PR #2328, opened 2019-06-18
edge case Gen3 PIDType

pokecal:
in Emerald, I have got ACDE pattern of PID type.

### DIFF
--- a/PKHeX.Core/Legality/RNG/MethodFinder.cs
+++ b/PKHeX.Core/Legality/RNG/MethodFinder.cs
@@ -83,6 +83,40 @@ namespace PKHeX.Core
             var reg = GetSeedsFromPID(RNG.LCRNG, top, bot);
             var iv1 = GetIVChunk(IVs, 0);
             var iv2 = GetIVChunk(IVs, 3);
+            if (GetLCRNGMatchMain(reg, iv1, iv2, out pidiv)) return true;
+            reg = GetSeedsFromPIDSkip(RNG.LCRNG, top, bot);
+            if (GetLCRNGMatchMain(reg, iv1, iv2, out pidiv))
+            {
+                switch (pidiv.Type)
+                {
+                    case PIDType.Method_2: pidiv.Type = PIDType.Method_2; break; // ACDE
+                }
+                return true;
+            }
+            return GetNonMatch(out pidiv);
+        }
+
+        private static bool GetLCRNGUnownMatch(uint top, uint bot, uint[] IVs, out PIDIV pidiv)
+        {
+            // this is an exact copy of LCRNG 1,2,4 matching, except the PID has its halves switched (BACD, BADE, BACE)
+            var reg = GetSeedsFromPID(RNG.LCRNG, bot, top); // reversed!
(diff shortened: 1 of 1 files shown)


kwsch:
Was the pkm obtained on an emulator?

pokecal:
GBA SP and a cartridge(Japanese), I used.

pokecal:
PID=3dd1bb49, IVs=(23,12,31,9,3,3)=(HP,Atk,Def,Speed,SpAtk,SpDef)
this only match to:
seed = 61c2b9f4
A = bb49e557, PID_low=bb49
B = baa1887e
C = 3dd1e219, PID_high=3dd1
D = 7d974318, IVs=(23,12,31)
E = 8c6941ab, IVs=(9,3,3)
[327 - パッチール - 88283DD1BB49.zip](https://github.com/kwsch/PKHeX/files/3305853/327.-.-.88283DD1BB49.zip)

kamronbatman:
What kind of encounter?

kwsch:
Better known as "Method 3", this PIDIV split occurs when the vblank interrupt occurs during PID creation.

It's more often seen in between the IVs or right after the PID, which is due to the game's overuse of a slow modulo function and tying the game loop with the display (vsync).
[More info here](https://www.smogon.com/forums/threads/rng-manipulation-in-firered-leafgreen-wild-pok%C3%A9mon-supported-in-rng-reporter-9-93.62357/post-3807320)

For the past decade or so, we've essentially ignored Method 3's existence as it is statistically unlikely: the slow modulo function consumes much more computation time.

During encounter generation, the game has to calculate random data for the new encounter. For wild Pokemon, we have to check if the encounter happens, what encounter slot, what level, what nature, and then we have to get a valid PID. This requires creating PIDs until a satisfactory value is encountered. Then, we calculate the IVs, and start the battle. (Held item is somewhere in here).

Anyways, for every random value, we're doing a modulo to reduce it to the required range. IVs are set after initial pkm data population, so if the timing was slow enough it happens there. Overall, these operations are an order of magnitude slower than getting two random numbers sequentially.

Emulators can have different speeds than the original hardware, which is why I asked. Although I haven't emulated the entire generation sequence to see the assembly instruction counts/speed for each step, I assume it's rarer to encounter a Method 3 spread than a shiny :)

I probably will come up with a similar solution to what you have, but will place the PIDIV type check at the end since it's incredibly uncommon. Leaving the pull request open until then :+1:

pokecal:
I have used nocasheGBA and read the encounter generations. I suspected the cost of supporting  the method, but it was so optimistic.
I want to avoid too, the big cost and less profit.

kamronbatman:
@kwsch, who should do the next step in this PR? Will you be using this PR to make a new one that is optimized?

kwsch:
@kamronbatman yes I'll do it later today (in the next 8 hours or so).

kamronbatman:
My life is complete lol

pokecal:
Thanks, and I have got correct knowledge about the interrupts, and difficulty of checking stat legality that use randoms.

PR #2342, opened 2019-06-28
Update GO Shiny banlist for GO 3rd Anniversary

sora10pls:
All Shiny Alola Forms are releasing; modify Verifier to allow Shiny Alola Form while still flagging the respective Kanto Form (if not released).

Changes go live at 20:00 UTC today (in just over 2 hours), getting it out of the way now.

image
image


kwsch:
Galarian forms might require a re-evaluation of this logic; tagging with galar keywords as something to look out for ~6 months from now.

eg if dratini gets a galarian form, another branch of code would have to check for nonshiny|shiny similar to what used to be done for alola.

PR #2344, opened 2019-07-01
update MailBoxEditor

pokecal:
fixed FRLG offset, etc.


pokecal:
I'm now trying to apply "Style updates"-like styles, and add another fix.

PR #2362, opened 2019-07-23
FR/LG flag names

rboninsegna:


kwsch:
Extra flags for reference:
https://github.com/pret/pokefirered/blob/e4acfdf20b7d56b83cee6f4e49b257a2ecb92a95/include/constants/flags.h#L1399

Names were originally copied from the pret repos ;)

Issue #2198, opened 2018-12-05
Build with .NET Core 3


rogersachan:
Winforms/WPF is now open source and .NET Core includes support for them according to <https://blogs.msdn.microsoft.com/dotnet/2018/12/04/announcing-net-core-3-preview-1-and-open-sourcing-windows-desktop-frameworks/> . What effort needs to be done in order to compile this application under .NET Core 3?

kwsch:
I've kept dependencies of PKHeX.WinForms to a minimum; only dependencies are:

- QRCoder, which isn't GUI and is only tied to System.Drawing
- pokesprite, which is only tied to System.Drawing like all other image resources.

There's one pinvoke done for windows builds to load the gen6 font file (see `FontUtil.cs`), which may be the only problem in the codebase that would need to be reconfigured. It's the only reason there's a mono build; no other special considerations :)

As for building, I wasn't able to figure it out yet -- I installed VS 2019 preview and the 3.0 sdk&runtime and couldn't change the build target for PKHeX.WinForms from .NET Framework 4.6 to .NET Core 3 (didn't show up in the list). Maybe need to wait for some dust to settle before a clear path is shown.

@evandixon migrated PKHeX.Core to netcore previously; maybe he would have an idea ;)

evandixon:
> As for building, I wasn't able to figure it out yet -- I installed VS 2019 preview and the 3.0 sdk&runtime and couldn't change the build target for PKHeX.WinForms from .NET Framework 4.6 to .NET Core 3 (didn't show up in the list). Maybe need to wait for some dust to settle before a clear path is shown.

> @evandixon migrated PKHeX.Core to netcore previously; maybe he would have an idea ;)

.Net Core uses a new project type, and while I absolutely adore it, I don't know of a way to move from the old to the new without creating a brand new project, copying the files over, and fixing whatever syntax errors there may be.

One thing to watch out for is the text resources. .Net Core 1.x and 2.x don't seem to play nice with .resx files having text files that are treated as strings, hence [some stuff I had to do in DataUtil that feels bad](https://github.com/kwsch/PKHeX/blob/master/PKHeX.Core/Util/DataUtil.cs#L131).

> I've kept dependencies of PKHeX.WinForms to a minimum; only dependencies are:
> QRCoder, which isn't GUI and is only tied to System.Drawing
> pokesprite, which is only tied to System.Drawing like all other image resources.

[...]

PR #2391, opened 2019-09-23
Initial switch to .Net Core 3.0

evandixon:
For the most part, this just worked. And that makes me skeptical. But this should be good enough to fix issues as they're found.

Part of #2198, if not the whole thing.


PKHeX.WinForms/App.config
```diff
@@ -1,84 +0,0 @@
-<?xml version="1.0"?>
-<configuration>
-    <configSections>
-        <sectionGroup name="userSettings" type="System.Configuration.UserSettingsGroup, System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" >
-            <section name="PKHeX.WinForms.Properties.Settings" type="System.Configuration.ClientSettingsSection, System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" allowExeDefinition="MachineToLocalUser" requirePermission="false" />
-        </sectionGroup>
-    </configSections>
-<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6"/></startup><userSettings>
```
evandixon: I guess I didn't preserve settings

evandixon:
Shouldn't have removed PKHeX.WinForms/icon.ico

PKHeX.WinForms/App.config
evandixon: They seem to still work 🤷‍♂️

evandixon:
Automated builds are failing because .Net Core 3.0 is too new. I've paused all the Team City build agents. Going forward, I think Azure DevOps would be a better choice, since they have free build agents that I don't have to maintain myself, but I'm open to discussion.

Issue #557, opened 2016-12-02
Interface problems: some icons and background seen corrupted


arturfelipet:
After an GitHub resync and compile this week this interfaces erros showed up for me, tried delete and re download the master and recompile and same result.

The transparent background shows black, the icons of file menu looks corrupted and the new icons for marking in gen 7 started showing bad too.

any ideas how do i fix this ? i use xamarin 6.1.2 build 44 to compile it on mac os 10.12.1 ... thanks.

![background errors 1](https://cloud.githubusercontent.com/assets/515567/20846866/c6c9a59e-b8b2-11e6-896a-3d1f5fef6b6a.png)
![background errors 2](https://cloud.githubusercontent.com/assets/515567/20846868/c8c32230-b8b2-11e6-8274-f7a251c9fd87.png)

kwsch:
Gonna blame https://github.com/kwsch/PKHeX/pull/517, looks like the optimization did something to the transparency data.

PR #2402, opened 2019-10-16
Optimize all png images in PKHeX.Drawing

ndye:


kwsch:
Related: https://github.com/kwsch/PKHeX/pull/517

kwsch:
Can you provide any details on what tools/settings were used to optimize the png compression?

I don't see any issue with the sprites on my windows machine, but due to prior attempts by others being unsuccessful for running under mono/wine, I don't feel it is appropriate to merge until someone can confirm this commit doesn't result in the same error as seen in #557.

ndye:
Sure, so I actually run on macOS so I can confirm that it runs on mono/wine and as for the settings a actually used opting and if you have ever used it it essentially attempts all of the possible compression options for the file and chooses the best one. if you are interested the actual command used to compress them is:
`optipng -o7 -zm1-9 -strip all -clobber -fix *.png`

kwsch:
Cool; can you confirm that hovering over the slots produces the intended glow effect, and that the sprites retain their original transparency appearance (see the readme for the expected view, and compare it to the referenced issue where the sprite turned black)

PR #2461, opened 2019-11-19
Connecting Areas Check

ReignOfComputer:
Like LGPE, SWSH's Wild Area has Pokemon that can bleed over routes. I've checked areas where this can happen and have made a method that returns other areas the Pokemon could have originated from.


ReignOfComputer:
Whoops - this seems to have broken Legality for random other things in the Wild Area. Any ideas?

PR #2484, opened 2019-11-21
Add another Lapras Spawn

Bappsack:
Lapras Wild Encounter at North Lake Miloch (in a Wild Area)

[131 - Lapras - E08F0BBE8452.zip](https://github.com/kwsch/PKHeX/files/3873293/131.-.Lapras.-.E08F0BBE8452.zip)


ReignOfComputer:
@kwsch is the list a definite list as it is?

kwsch:
The list is definite; no other spawns are available.
Once you beat the game, all levels are bumped to 60, so this is just a misidentified encounter.

Bappsack:
i hightly doubt the list is definite, i got some Pokemon with Encounters that aren't exist in Encounters8.cs, they also don't have crossover locations, check the attached Lapras.

at North Lake Miloch (in a Wild Area) can't crossover with at the Lake of Outrage (in a Wild Area).

SciresM:
@Bappsack The list we have is dumped directly from deserialized gamedata.

Bappsack:
@SciresM i know that, but how i get the Lapras there then?

SciresM:
Level 60 --> scaled up by wild area. Presumably it's some other lapras.

One way to test would be to get the same encounter on a save without all the badges, to see what the encounter's actual level is.

Bappsack:
asuming its the same level, 56 before beating the game since u need to be able to surf, but in this case thats not the point, the Lapras is from a different Area, not Sea of Outrage, Kurts documentations about Lapras doesnt say a Lapras spawns there.

PR #2489, opened 2019-11-22
add a check for gigantamax flag during encounter match

architdate:
Doesnt let non gmax flag encounters match when CanGigantamax is true


architdate:
Don't merge yet, fixing some more Nest related stuff!

PR #2490, opened 2019-11-22
Misc Nest Fixes

architdate:
Separate PR from the one I posted since you can separately merge this if you have an alternate solution for #2489 
This change is based on a lot of experimentation done in my discord.


kwsch:
Need to move the ability check to the deferral check instead of wholly rejecting the encounter.
I'll handle it in a future commit.

PR #2643, opened 2020-01-18
Add Subway Scores edition

egzonqj:
Adding some initial changes to edit values of the Subway Scores. Currently only available for Black and White, I still have to figure out if the offsets are the same at all on B2W2...

image

This could have been coded better tho...

### DIFF
--- a/PKHeX.Core/Saves/Substructures/Gen5/BattleSubway5.cs
+++ b/PKHeX.Core/Saves/Substructures/Gen5/BattleSubway5.cs
@@ -7,29 +7,6 @@ namespace PKHeX.Core
         public BattleSubway5(SAV5BW sav, int offset) : base(sav) => Offset = offset;
         public BattleSubway5(SAV5B2W2 sav, int offset) : base(sav) => Offset = offset;
 
-        private const int superCheckOffset = 0x21D04;
-
-        private const int singlePasOffset = 0x21D08;
-        private const int singleRecOffset = 0x21D1A;
-        private const int doublePasOffset = 0x21D0A;
-        private const int doubleRecOffset = 0x21D1C;
-        private const int multiNpcPasOffset = 0x21D0C;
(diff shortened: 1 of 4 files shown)


kwsch:
Block Offsets for B2/W2:
https://github.com/kwsch/PKHeX/blob/f3a9a9d7c2f80ed7331af78c1615de5bb88094eb/PKHeX.Core/Saves/Access/SaveBlockAccessor5B2W2.cs#L65-L67

Block Offsets for B/W:
https://github.com/kwsch/PKHeX/blob/f3a9a9d7c2f80ed7331af78c1615de5bb88094eb/PKHeX.Core/Saves/Access/SaveBlockAccessor5BW.cs#L67-L69

These should be kept in separate blocks.

The offsets should be the same regardless of version, seeing as the data size of each block doesn't change. For this, just subtract the raw block offset from the const one you have, then inline it.

Example:

```cs
private const int superCheckOffset = 0x21D04;
```
```cs
        public int SuperCheck
        {
            get => BitConverter.ToUInt16(Data, superCheckOffset);
            set => BitConverter.GetBytes((ushort)value).CopyTo(Data, superCheckOffset);
        }
```
Since we know that the block data starts at 0x21D00, this becomes:
```cs
        public int SuperCheck
        {
            get => BitConverter.ToUInt16(Data, Offset + 0x04);
            set => BitConverter.GetBytes((ushort)value).CopyTo(Data, Offset + 0x04);
        }
```

When we create the block, we already pass in the block offset (0x21D00 for B2W2, and 0x21B00 for B/W), so it'll innately support both versions with the `Offset + ofs` style of read/write used for the other blocks.

egzonqj:
Ah I see, thanks for the explanation! Hope it's all good now. Since I didn't properly check the Accessors, I went with the BW-only approach at that time, apologies.

PR #2650, opened 2020-01-21
set gen 5 battle subway record max to 9999 + tab indexing

murrty:
since battle subway records has a maximum of 9999 for your record. non-super formats don't really need a maximum, but i included it anyways.

i also included some tab indexing so people can navigate the window with a keyboard, and a minor fix to the "record" labels so you can read them.


murrty:
i didn't realize how much the damn auto-generator for visual studio changes stuff for minor changes

PR #2686, opened 2020-02-06
RaidSpawnList8 fix DenType and Flags

3096:


kwsch:
The changing of type might break the Raid Plugin until the next release; holding off until then (probably tomorrow)

slp32:
Shouldn't `HasWatts` be called `HasHarvestedWatts`? Since that's what it means if it's set.

By the way @kwsch, could you make it so Den 017 is not edited when clicking Activate All? That is the den for Watchtower Lair which only has encounters if you use a Dynamax Crystal, and setting it to Rare doesn't do anything. (it might use an undocumented den type/flag?)

3096:
> Shouldn't `HasWatts` be called `HasHarvestedWatts`? Since that's what it means if it's set.

You're right. That's a mistake on my part. Let me redo the commit as well as get rid of the nested ternaries.

3096:
Well, technically the game seems to use `RaidType.Event` for non wishing piece event dens, so I don't think `Common` + event flag is legal. I don't think `Rare`/`RareWish` + event flag combinations are possible either (for now).
The game seems to handle any combination just fine tho, so I didn't feel like writing checks for everything because the logic gets convoluted (thank GF for not making up their mind on what they want to use lmao.)
Notably  `RaidType.Event` has no impact whether if a den is actually treated as event or not, and it's solely determined by the flag, so I think I'll add a few simple checks to avoid confusion.

PR #2694, opened 2020-02-10
Add Gender to Gen 8 Trainer Card

CanoeHope:
In other interesting news, a Block file with all available clothing was posted on PP!: https://projectpokemon.org/home/files/file/4041-fashion-block/

Maybe it could be implemented into PKHeX? I'm sure a lot of peps would love this! :)


kwsch:
I'd rather not embed blocks; ppl can just grab them and import like they can for the download raid news.

CanoeHope:
Oh, I thought it would be akin to SM/USUM where there was a button in the trainer editor to give yourself all clothing.

Thanks for your input though!

PR #2700, opened 2020-02-11
Change FRLG hatch location

CanoeHope:
Better location.


kwsch:
The basis for the original locations was just to be closest to the daycare, rather than the most optimal hatch-counter straight-line location ;)

PR #2750, opened 2020-03-05
Update LegalityCheckStrings_es.txt

egzn:
Wanted to review Phoenix's changes but figured out this would be better. Not putting away their efforts on the translation tho, I'm also much appreciated for it!


XxPhoenix1996xX:
It does look better, thank you!

PR #2760, opened 2020-03-13
Simplify foreach loops

pi1024e:
In these loops, the var is picked out of the value and then determined
if it satisfies a condition or not. However, why not use the built-in
functions to determine the compliant values and operate on those? It
even looks better!

### DIFF
--- a/PKHeX.Core/Util/ArrayUtil.cs
+++ b/PKHeX.Core/Util/ArrayUtil.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace PKHeX.Core
 {
@@ -163,9 +164,7 @@ namespace PKHeX.Core
 
         internal static T[] ConcatAll<T>(params T[][] arr)
         {
-            int len = 0;
-            foreach (var a in arr)
-                len += a.Length;
+            int len = arr.Sum(a => a.Length);
 
             var result = new T[len];
 
--- a/PKHeX.Core/Util/DataUtil.cs
+++ b/PKHeX.Core/Util/DataUtil.cs
@@ -333,8 +333,7 @@ namespace PKHeX.Core
         public static List<ComboItem> GetCBList(IReadOnlyList<string> inStrings, IReadOnlyList<ushort> allowed)
         {
             var list = new List<ComboItem>(allowed.Count + 1) { new ComboItem(inStrings[0], 0) };
-            foreach (var index in allowed)
-                list.Add(new ComboItem(inStrings[index], index));
+            list.AddRange(allowed.Select(index => new ComboItem(inStrings[index], index)));
             list.Sort(Comparer);
             return list;
         }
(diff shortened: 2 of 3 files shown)


kwsch:
The goal for these is to avoid using linq extensions when they aren't explicitly needed, to avoid any unnecessary allocations in methods that are called frequently/on init.

Example: The ConcatAll method:

Original (without LINQ):
```il
ConcatAll2<T>:
IL_0000:  ldc.i4.0    
IL_0001:  stloc.0     // len
IL_0002:  ldarg.0     
IL_0003:  stloc.1     
IL_0004:  ldc.i4.0    
IL_0005:  stloc.2     
IL_0006:  br.s        IL_0016
IL_0008:  ldloc.1     
IL_0009:  ldloc.2     
IL_000A:  ldelem.ref  
IL_000B:  stloc.3     // a
IL_000C:  ldloc.0     // len
IL_000D:  ldloc.3     // a
IL_000E:  ldlen       
IL_000F:  conv.i4     
IL_0010:  add         
IL_0011:  stloc.0     // len
IL_0012:  ldloc.2     
IL_0013:  ldc.i4.1    
IL_0014:  add         
IL_0015:  stloc.2     
IL_0016:  ldloc.2     
IL_0017:  ldloc.1     
IL_0018:  ldlen       
IL_0019:  conv.i4     
IL_001A:  blt.s       IL_0008
IL_001C:  ldloc.0     // len
IL_001D:  newarr      UserQuery.T
IL_0022:  ret         
```
With LINQ:
```il
ConcatAll<T>:
IL_0000:  ldarg.0     
IL_0001:  ldsfld      UserQuery+<>c__2<>.<>9__2_0
IL_0006:  dup         
IL_0007:  brtrue.s    IL_0020
IL_0009:  pop         
IL_000A:  ldsfld      UserQuery+<>c__2<>.<>9
IL_000F:  ldftn       UserQuery+<>c__2<>.<ConcatAll>b__2_0
IL_0015:  newobj      System.Func<,System.Int32>..ctor
IL_001A:  dup         
IL_001B:  stsfld      UserQuery+<>c__2<>.<>9__2_0
IL_0020:  call        System.Linq.Enumerable.Sum<T[]>
IL_0025:  newarr      UserQuery.T
IL_002A:  ret         

<>c__2`1.<ConcatAll>b__2_0:
IL_0000:  ldarg.1     
IL_0001:  ldlen       
IL_0002:  conv.i4     
IL_0003:  ret         

<>c__2`1..cctor:
IL_0000:  newobj      UserQuery+<>c__2<>..ctor
IL_0005:  stsfld      UserQuery+<>c__2<>.<>9
IL_000A:  ret         

<>c__2`1..ctor:
IL_0000:  ldarg.0     
IL_0001:  call        System.Object..ctor
IL_0006:  ret    
```

The LINQ rewrite generates more/larger IL, and has newobj allocation for the user query AND the func selector.

=========

Sure, it looks cleaner, but we don't need the extra weight that LINQ brings for these heavily used methods.

PR #2765, opened 2020-03-15
Made the "Contains" function non-recursive.

pi1024e:
So far, if g1 is SWSH, the function will recurse. Since SWSH is one of the most highest selling pokemon games, it is very likely g1 will be SWSH.

To address this, I made the statement iterative instead of recursive. I hope this helps

### DIFF
--- a/PKHeX.Core/Game/GameUtil.cs
+++ b/PKHeX.Core/Game/GameUtil.cs
@@ -174,76 +174,96 @@ namespace PKHeX.Core
         /// <param name="g2">Individual version</param>
         public static bool Contains(this GameVersion g1, GameVersion g2)
         {
-            if (g1 == g2 || g1 == Any)
+            if (g1 == g2 || g1 == Any) 
                 return true;
 
-            switch (g1)
+            do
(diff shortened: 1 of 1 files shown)


kwsch:
Resharper can offer a lot of neat refactorings, but please don't blindly apply them without considering the ramifications.

* This doesn't make it non-recursive; it only loops if the input is Gen8, and every other case remains recursive (using different arguments).
* The maximum recursion depth of this method is only 2; the lumped versions will only recurse once.
* There isn't a single usage for the Gen8 enum member calling `Contains`; it's only used as a unique for other lookup tables / descriptors for move origin. The code that you've changed is never reached with the current program logic, but still remains for others to use via API calls / future updates (new games).
* If the DPPt remake is next year, Gen8 will have to have to have another case for that other lumping version, which can't be done by a while loop (needs to recurse with different arguments). The changes would have to be undone, and it'd have to look like the other branches (Gen7) to compare multiple.
* This adds 20 lines because it didn't keep the case->return on the same line like the original code

PR #2766, opened 2020-03-15
Use switch expressions instead of switch statements

pi1024e:
This is so we can make the code more consistent across files.

If there are any changes you would like to suggest, please let me know.

Thank you!


kwsch:
Although switch expressions can be nice, they do not allow setting breakpoints for individual cases. This isn't any real value added. If anything, it introduces more noise as some branches were duplicate, resulting in duplicated code.

I have numerous refactoring extensions installed in my setup; things are already quite polished as they are. **Please don't submit any more pull requests for cosmetic language tweaks.**

PR #2834, opened 2020-05-12
Add flag for Foreign DexNav Encounters Unlocked

ultima-soul:


sora10pls:
Out of curiosity, is this the only notable flag that’s set after the Kyogre/Groudon climax of the games? It may be worth rewording the flag if so, since several things only open up after the fact.

Foreign DexNav, National Pokédex becoming available, remainder of the Mega Stones, sidequests like Sea Mauville/New Mauville, etc.

ultima-soul:
These are the constant and flag differences I got right before Kyogre battle and right after it when I could save again.

```
33: 178->0
49: 0->4136
69: 14->20
112: 6->8
147: 1->8
166: 5->7
170: 3->5
173: 1->2
237: 0->1
245: 0->2

Flags set:
0205 Nope, 0453 Nope, 0458 Nope, 0647 Kyogre Defeated, 0656 Nope, 0662 Nope, 0663 Nope, 0664 Nope, 0665, 0666, 0667, 0680, 0937 Nope, 2630 Nope, 2642 Yes, 2927

Flags unset:
0414, 0432, 0442, 0655, 0659, 0661, 0920, 0959, 0960, 0961, 0962, 0963, 0964, 0965, 0967, 0969, 0970, 0972, 0974, 0975, 0977, 2810
```
I specifically tested unsetting each of the flags that were set, one-by-one, on a different completed save file after getting this diff to see if I could revert the unlock. The ones that were "Nope" didn't affect the foreign DexNav encounters from being displayed on the DexNav screen specifically. The one that was Yes hid the foreign encounters and downgraded the completion icon to the Gold one where you've caught all of the Hoenn Dex mons in the area. Setting it afterwards displays them again and the Platinum icon comes back. 

Afterwards, I tested it on a new save file to see if you could get them as soon as you can use the DexNav. Once you've completed the Poochyena static encounter and then talked to Brendan on Route 102 where the Beautifly fly by, you can then get random DexNav mons popping up. And indeed, the foreign DexNav mons started appearing.

sora10pls:
Thank you for the list of flags! Some time in the future I can do some testing to see which flags would be nice to add to the current list, like stuff I mentioned above.

PR #2841, opened 2020-05-25
Restrict the Excluded Pokemon in IDynamaxLevel to between Zacian and Eternatus Inclusive

foohyfooh:
Preemptively restrict the Dynamax Level check to between Zacian and Eternatus inclusive since the DLC trailer depict Urshifu as Dynamaxed.


kwsch:
per my conversations with @sora10pls 

image

They'll likely add new legendaries which can't be dynamaxed. It's on our list of things to check when the DLC comes, prior to any update. We don't want to be too proactive with assumptions at this time.

PR #2843, opened 2020-05-28
Gen3 flags rs translate

kamisama6866:
[flags_rs_zh.txt](https://github.com/kwsch/PKHeX/files/4694988/flags_rs_zh.txt)


kamisama6866:
I'm not very good at using GitHub..

PR #2878, opened 2020-06-20
Update legal raid counts

ca1e:
the normal raids have 100 before and the raids count of armor area should be 90?


kwsch:
I must have confused the raw nest list from the actual amount of interactable dens on the map (which can choose from the raw nest list).

PR #2879, opened 2020-06-20
Unblock IoA Mons Auto Dynamax

ReignOfComputer:
Probably easier to blacklist this way?


sora10pls:
Should do:  >= Zacian && <= Eternatus
Only disallows 888-890.

architdate:
Its the same thing that ROC posted. Its just easier to understand ROCs code. The actual implementation for future proofing should be an array of species in `Tables.cs` with list of undynamaxable species and then check if `array.Contains(species)`

Issue #2886, opened 2020-06-20
the button of "randomize ivs"  has something wrong


gjfkkk:
**Describe the bug**
the button of "randomize ivs" can be used(at least 3 ivs=31) when the species are 887 888 890 but not 880 881 882 883

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

ReignOfComputer:
For clarity: Looks like the Fossil Pokemon are also 3IV minimum, and PKHeX doesn't take that into account. Can reproduce.

kwsch:
Not an issue. The randomization of IVs is unrelated to encounter IV requirements.

PR #2887, opened 2020-06-20
3IV Galar Fossils

ReignOfComputer:
Fixes https://github.com/kwsch/PKHeX/issues/2886. Would like additional confirmation from others that the Fossils are indeed 3IV, but from my own checks all my Fossils are min-3IV.


kwsch:
They aren't classified as sublegends; the encounter detail just enforces 3IV.

The IV randomization is "dumb" in a sense, where it doesn't refer to the encounter details, only legend status. XY has a special condition for unbreedables, but I don't think adding in more complexity is worth it. The user just needs to know that some encounters require certain IVs.

Closes #2886

sora10pls:
This isn’t the ideal solution imo. There are dozens of static encounters, gifts, and ingame trades that are given a set number of flawless IVs (usually 3 or 6). When randomizing IVs, it should check for the given encounter’s FlawlessIVCount.

Galar Fossils are not Sub-Legendary.

PR #2968, opened 2020-08-14
Changes to suggested hyper training logic 

rat01:
Now that Hidden Power is removed from the game in gen 8, IVs have no battle relevant side effects and only affect the pokemon's stats. This means that any IVs specified in the showdown format are deliberate for stat purposes and should not be hyper trained over.

The main use case for this is allowing sysbots to gen trick room pokemon with a specific amount of speed creep without bleeding into another speed tier.

### DIFF
--- a/PKHeX.Core/Editing/CommonEdits.cs
+++ b/PKHeX.Core/Editing/CommonEdits.cs
@@ -202,7 +202,11 @@ namespace PKHeX.Core
             pk.IVs = Set.IVs;
             pk.EVs = Set.EVs;
 
-            pk.SetSuggestedHyperTrainingData(Set.IVs);
+            // IVs have no side effects such as hidden power type in gen 8
+            // therefore all specified IVs are deliberate and should not be HT'd over for pokemon met in gen 8
+            if (!pk.Gen8)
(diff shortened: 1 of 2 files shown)


kwsch:
I don't think this interacts well with past-gen exclusive learnsets, where flawless IVs might not be possible for a given nature.

rat01:
I was hesitant to change the hyper training logic at first because I was worried about causing unintended behavior with the mass hyper train box feature, but that solution you suggested sounds like it should cover that too. I'll change my commit to use this solution later today

rat01:
The hyper training suggestion logic still doesn't hyper train the speed IV if it's 31 or less than 2, like before, but now it also doesn't hyper train if no other stats are hyper trained and the IV is less than 17. I picked 17 because it covers about half the range, and 16 was the highest I needed in any specific case of doing this.
image

In most cases, this works the same as before.
image
image

I also kept something similar to what I had before, except showdown import won't attempt to hyper train any pokemon __encountered__ in gen 8 instead of all gen 8 format pokemon. Same reasoning as before, but no clashes with old generation stuff.
image

kwsch:
Thanks :)

PR #2974, opened 2020-08-19
Refactoring some IEncounterable-type'd object initialization and original-generation evo chain

kwsch:


PKHeX.Core/Legality/Areas/EncounterArea7.cs
```diff
@@ -8,74 +8,36 @@ namespace PKHeX.Core
...
     {
-        protected override IEnumerable<EncounterSlot> GetFilteredSlots(PKM pkm, IEnumerable<EncounterSlot> slots, int minLevel)
+        public override IEnumerable<EncounterSlot> GetMatchingSlots(PKM pkm, IReadOnlyList<EvoCriteria> chain)
         {
-            int species = pkm.Species;
-            int form = pkm.AltForm;
-
             // Edge Case Handling
-            switch (species)
+            switch (pkm.Species)
```
kwsch, line 14: This edge case handling might not be necessary anymore due to the cleaned form matching. To be tested later.

PR #2977, opened 2020-08-25
Handle Exception When Adding Font File

foohyfooh:
It appears that some users are experiencing an issue where the PrivateFontCollection cannot find the font that should have been written to the temporary path. Move the AddFontFile call into the try-catch block so that PKHex doesn't crash in this event.

Link to Issue: https://projectpokemon.org/home/forums/topic/57234-pkhex-200414-error/


foohyfooh:
Link to other instance of the issue: https://projectpokemon.org/home/forums/topic/57741-pkhex-worked-the-1st-time-i-opened-it-now-it-doesnt-work-at-all/

PR #2987, opened 2020-09-13
Add support for reading and editing Underground Items

egzn:
This PR brings support to view and modify the Underground Items from any Sinnoh save. Months ago I did some [research on the offsets](https://projectpokemon.org/home/forums/topic/56097-dppt-underground-offsets-for-items/) for these items in order to try and add support when I could (or for others to do it, of course).

image

Comparison to PokeSAV's: 
image

Any changes/suggestions are very welcome.


kwsch:
Thanks for implementing this! <3

PR #2989, opened 2020-09-14
Fix Hidden Power Power calculations (and add to UI)

soopercool101:
The power calculations for Hidden Power used the incorrect bit. Fixed for both Gen II and Gen III-V. Added Power calculations to UI for Gens 1-5 where they are applicable

### DIFF
--- a/PKHeX.Core/PKM/PKM.cs
+++ b/PKHeX.Core/PKM/PKM.cs
@@ -491,15 +491,19 @@ namespace PKHeX.Core
 
         protected static int GetHiddenPowerBitVal(int[] ivs)
         {
+            return GetHiddenPowerBitVal(ivs, 0);
+        }
+        protected static int GetHiddenPowerBitVal(int[] ivs, int bit)
+        {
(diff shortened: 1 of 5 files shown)


soopercool101:
Marking as a draft because, despite the math seeming to match up, the values for Generation I/II don't appear to line up perfectly with http://www.psypokes.com/gsc/dvguide.php likely due to rounding differences?

soopercool101:
The rounding errors are on the site, not on the current code

kwsch:
Disassembly:

Crystal:
https://github.com/pret/pokecrystal/blob/3eacab563d0e1ab5557c2443556a7a5e58d14cad/engine/battle/hidden_power.asm

Ruby (Gen3+)
https://github.com/pret/pokeruby/blob/75cea2933f50a96052bc3e498a66dfe6401d8ca0/src/battle_script_commands.c#L12497-L12521

soopercool101:
Matches in-game and disassembly values for both from all my tests👍

PR #3041, opened 2020-10-23
Update text_Items_zh.txt

CScorpion-h:
Match Crown Tundra


kwsch:
This will be addressed in the cumulative update coming later for all languages. Help isn't needed at this time!

PR #3076, opened 2020-11-04
update chinese translation

mickeyandkaka:


kwsch:
Thanks for the contribution; I went with @easyworld 's translations instead as it had more strings translated and it had updated the punctuation characters.

If there's further translations needed, please raise another issue/pull request :)

PR #3080, opened 2020-11-05
Added Bill's Eevee Gift and Route 36 Sudowoodo flags in HGSS

ShadyRelapse-PP:
Just added flags


ShadyRelapse-PP:
Just edited the names to make them fit more.

PR #3116, opened 2021-01-02
Added Bebe's PC Event Flag for DP + Update Gen 3 & 4 translations

CanoeHope:
I also updated the Spanish FRLG translation file with 2 Flags.


sora10pls:
You might want to add these flags to other languages’ resources (this includes those from your last PR). They can be translated by others at some other point, but it’s better to not make their visibility exclusive to English users.

CanoeHope:
> You might want to add these flags to other languages’ resources (this includes those from your last PR). They can be translated by others at some other point, but it’s better to not make their visibility exclusive to English users.

Just finished doing so. Thanks for the heads up!

PR #3141, opened 2021-02-07
Correct HOME Shiny Zeraora date

CanoeHope:
https://bulbapedia.bulbagarden.net/wiki/List_of_game-based_Pok%C3%A9mon_distributions_in_Generation_VIII#1_Million_Victories_Shiny_Zeraora


Lusamine:
HOME time is based on UTC.  30th should be the correct date since 5 PM PST on the 29th is 1 AM UTC on the 30th.

CanoeHope:
But my legitimately claimed one tripped the legality checker and it in fact has 29th as the date (I'm in EST time, btw).

Issue #2916, opened 2020-07-01
request : small items lower on box slot


poke-hexedit:
**Is your feature request related to a problem? Please describe.**
items are sometimes too high when in view of a pokemon slot .

**Describe the solution you'd like**
if its possible i would like if smaller items can be put closer to the edge of the box slot like some big items fit fine but if a pokemon has for example a mega stone then it is very high and not close to border . for example in this the candy should be closer to bottom
![image](https://user-images.githubusercontent.com/35694052/86285972-9b768f80-bbb3-11ea-9b1f-52fa2d515a44.png)

**Describe alternatives you've considered**
none

**Additional context**
i asked in the past about items being higher than they were but i think this would be a good change for smaller item pictures .

kwsch:
Closed by #3215 and #3213, thanks to @LukeThe 's initial PR and the follow-up item crop change.

PR #3213, opened 2021-06-08
[Enhancement#2916] Difference smalls items by null pixels

LukeThe:
Hello, this is my collaboration to PKHeX:
I've been working on the following issue: https://github.com/kwsch/PKHeX/issues/2916 . I suggested a feature that differences the small items's  sprites of the bigger ones by counting the blanks/nulls pixels. If a sprite has 800 or more nulls pixels it's a small one.
I've used the next source of topic-related information: http://csharphelper.com/blog/2016/10/count-pixels-of-different-colors-in-c/

Before:
Before

After:
After

**IMPORTANT**: The bigger sprites doesn't change their coordinates. The first column in the provided images is an example of small sprites and the second of biggers ones. 
You can modify the coordinates in the lines 140 and 141 of SpriteBuilder.cs

### DIFF
--- a/PKHeX.Drawing/Sprites/SpriteBuilder.cs
+++ b/PKHeX.Drawing/Sprites/SpriteBuilder.cs
@@ -131,12 +131,37 @@ namespace PKHeX.Drawing
                 _ => (Image?)Resources.ResourceManager.GetObject(GetItemResourceName(item)) ?? UnknownItem,
             };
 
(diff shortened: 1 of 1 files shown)


kwsch:
Closed by #3215 -- calling GetPixel->Name=="0" for each pixel is kinda slow & allocation heavy; I found it was better to just pre-crop the item sprites and shift them from the edge, resulting in all sprite sizes being at the bottom of the image.

PR #3215, opened 2021-06-11
Crop Item sprites, Layer with adjusted X coordinate

kwsch:
Before:
image

After -- Old Shift revised:
image
After -- New Shift:
image

Compiled Results:
dll: -20KB
Pixels to draw: -30%, by skipping transparent pixel rows

Summary:
All items are tucked closer to the pokemon, gravitating towards the bottom (2 pixel buffer at sides).

Further Improvements?
The shift-left `/4` can potentially be revised to be a smaller div so that smaller items are shifted left more (maybe `3.6f`?).


kwsch:
If this PR is merged, would address #3213 in a different way, closing #2916

sora10pls (approved):

PR #3221, opened 2021-06-20
Enforce weather legality for SWSH

Lusamine:
Eventual goal is to enforce weather legality in SWSH for available spawns and mark legality.

Creates a dictionary of possible weather types for each SWSH location based on [map weathers](https://docs.google.com/spreadsheets/d/e/2PACX-1vTFus5xQfPiswvc98vd5MlgHoWvdxSlprJyGCIEJKRoq3fKDZJnjZORctuHQvCkC47SLliGWSDP2RQj/pubhtml).  This is for weather legality for encounter slots, as static encounters have their permissible weathers denoted individually.  Unlisted locations may only have Normal weather or do not have any encounter slots.

There are edge cases where weather from area A can be used to spawn Pokémon in an adjacent area B where that weather does not occur normally.  Static encounters have these cases added to their individual entries.

Some but not all areas have symbol encounterslots that are affected differently from hidden slots.  

This includes pkl that removed inaccessible tables and with metadata to distinguish encounterslots.  Metadata for encounter slot type (e.g. symbol, hidden, surfing, sky, etc) are used to handle weather bleed cases.


PKHeX.Core/Legality/Verifiers/MarkVerifier.cs
```diff
@@ -69,9 +70,64 @@ public static bool IsMarkValid(RibbonIndex mark, PKM pk, IEncounterTemplate enc)
...
+                if (EncounterArea8.WeatherBleedHiddenGrass.TryGetValue(location, out var weather) && (weather.HasFlag(permit) || mainweather.HasFlag(permit)))
+                    return true;
+
+                // Allow all tree and fishing encounters.
+                weather = s.Weather;
+                if (weather.HasFlag(AreaWeather8.Shaking_Trees) || weather.HasFlag(AreaWeather8.Fishing))
+                    return true;
+            }
+
+            if (mainweather.HasFlag(permit))
```
kwsch, line 125: Not sure why this is checked last:
* Above branches check it when returning true
* Nothing returns false, so checking it first will do no harm?
* Should any of the inners return false?

If possible, check mainweather first, then we don't have to care about it after?

PR #3226, opened 2021-06-24
create a command line interface to support bots create pokemon from showdown files

Jungbluth11:
create a command line interface to support bots create pokemon from showdown files
commands:

- **loadsavefile "path/to/savefile"** // loads the savefile
- **loadshowdown "path/to/showdown"** // loads a file that containing showdown sets
- **createpkm** // creates PKM-objects from the showdown sets
- **setpkm showdownindex box slot** // sets a pokemon of the chosen showdown index to a specific box an slot
- **getshowdownsets** // lists the loaded showdownsets and shows the index
- **save "path/where/to/save"** // saves the savefile to the given path
- **exit** // closes the cli

### DIFF
--- a/PKHeX.CLI/App.config
+++ b/PKHeX.CLI/App.config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2" />
+    </startup>
+</configuration>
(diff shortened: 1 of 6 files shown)


kwsch:
[Auto Legality Mod](https://github.com/architdate/PKHeX-Plugins) already does this and properly modifies the data to match the Showdown Set, and anyone can wrap the functions in a separate CLI tool. SysBot.NET has an implementation that does automatic trading from Showdown Sets provided via discord/twitch.

This kind of thing should be a separate repo, not included in the main project. I don't think enough people would be using it to warrant it not being a separate project. PKHeX.Core is available on NuGet! Nice usage of the API though :)

PR #3254, opened 2021-09-03
Correct Gen 2 Snorlax Event Flag

CanoeHope:


kwsch:
Based on pret, it looks like the original snorlax flag is correct?

https://github.com/pret/pokecrystal/blob/2fe0cbbb19df504723934f39473064033c64ef6f/constants/event_flags.asm#L1326-L1327

Actually, you're correct, it uses a separate flag for the "vanish" portion:
https://github.com/pret/pokecrystal/blob/2fe0cbbb19df504723934f39473064033c64ef6f/constants/event_flags.asm#L1361

The original flag for Gyarados is correct (used as vanish) while the original snorlax one is used more for story stuff.

PR #3359, opened 2021-12-30
Generations evolution iteration

javierhimura:
The current PKHex Evolution Chain has a limitation for pokemon moved between generations because it allows chains in which the pokemon has existed on multiple generations after and before evolution, which is not possible (except Gen 1 and 2)

For example, a Gen 3 Butterfree with level 50 and meet level 10 on Platinum will have an Evolution Chain allowing the pokemon being Caterpie, Metapod, and Butterfree in both gen 3 and 4. That Evolution Chain is not possible, allowing incompatible moves like gen 3 Butterfree exclusive moves Double-Edge with gen 4 Metapod exclusive move Iron Defense. There are more pokemon with this kind of incompatible moves and it is not limited to gen 3 to 4.

Source
https://bulbapedia.bulbagarden.net/wiki/Butterfree_(Pok%C3%A9mon)/Generation_IV_learnset#By_leveling_up
It has moves in both "By transfer from another generation" that are exclusive to evolved Butterfree from gen 3 games and "By a prior evolution" that are exclusive from unevolved Metapod and Caterpie from gen4 games.

We don't know in which generation the pokemon has evolved so the only option is to obtain from the Evolution Chain the list of generations in which evolutions have been possible and iterate through it to obtain a reduced evolution chain that does not have evolutions in more than one generation. In the Buterfree example, there are 4 possible legal Evolution Chains based on the current Chain generated.

As a consequence of this change is no longer ambiguous in which generation the pokemon has evolved. This affects some checks like Shedinja moves, Feebas evolution, and evolution by move; those checks now use the generation in which the pokemon has evolved.

For the generation evolution iteration, both gens 1 and 2 are treated as just one generation because pokemon could be moved between both generations after and before evolve and any illegal move between gen 1 and 2 is already covered by other existing checks.


Lusamine:
I ran this over my collection and it's spitting out quite a few of these:
image
For example, a gen 3 Mew with tutor moves that was transferred to gen 5.
image
Here's the NPC trade.
image
image

Likewise, it's flagging all my Swords Dance Shedinja in gen 3.
image

There are also some edge cases we've had noted for a few months now that this currently doesn't handle:
- Level 1 Glaceon in SWSH with Aurora Veil.  It had to be a Glaceon to learn Aurora Veil via TM in gen 7, but SWSH requires it to have gained 1 level to be a Glaceon.
- Ruby Shedinja in SWSH with Sucker Punch and Hidden Ability. It had to be evolved in a past gen to learn Sucker Punch, and can't be Ability Patched in SWSH due to having the same ability in all 3 slots.

Here's a zip with some of these files.
[tests.zip](https://github.com/kwsch/PKHeX/files/7793882/tests.zip)

javierhimura:
I have fixed the error with Shedinja, I forgot to populate EvoGenerations lists for pokemon that don't require iterating through generations. I'll check the others later this night.

javierhimura:
I have fixed all the test pokemon except Level 1 Glaceon in SWSH and Ruby Shedinja in SWSH.

javierhimura:
Fixed Ruby Shedinja and similar pokemon. With this change, the gen 3 moves are marked as invalid with Hidden Ability because it iterates through all possible Evolution Chains keeping the last one (evolving in gen 8). I have avoided it by keeping the evolution chain that makes all moves valid but other data invalid and using that chain in case the encounter is invalid.

javierhimura:
Glaceon also fixed, not only for level 1 but also for Glaceon at encounter level and Glaceon at meet level originated from gens 1-4
I didn't realize there are more pokemon with evolutions that requires to gain a level before but not anymore in gen 8, I will improve it tomorrow to use a list of this type of evolved pokemon

Lusamine:
I ran the latest commit and I find this is incorrectly flagging Gen 3-4 Glaceon. Transfer from Gen 4 sets the met level to be the current level.

Other Pokemon that had an evolution method change from "evolve by leveling up at a location" to "evolve with a stone" are Leafeon and Magnemite in SWSH.  I've included a false negative Gen 7 level 20 Vikavolt in SWSH with Roost (Grubbin evolves to Charjabug at 20, so it couldn't have been a Vikavolt to be TM'd Roost).  There are probably other cases.

Of note, BDSP allows both Glaceon and Leafeon to be evolved with stones, but Ice Stone is not currently available. Milotic is also forced into a beauty requirement due to lack of Prism Scale.  Future inter-connectivity between the games will change some of these evolution restrictions.

[tests2.zip](https://github.com/kwsch/PKHeX/files/7795054/tests2.zip)

javierhimura:
Edge cases for Vikavolt fixed. It also applies to Magnezone. And probably to Probopass if is added into gen 8 in the future.

I also added edge cases for Magnezone from gen 1-4 evolved on gens 5 to 7, that pokemon should have one level more than met level.

Another edge case for Milotic evolved in gen 3 or 4, it should be one level more than the encounter.

That gen 3 Glaceon is legal because it could have been evolved in Sword/Shield with a stone. If you put a gen 7 exclusive move like Frost Breath it is flagged as illegal evolution because it should have leveled up to being able to evolve in USUM to learn that move.

There are existing checks in the master branch to not allow Glaceon and Leafeon stones evolutions nor Prim Scale Milotic for pokemon from BDSP and previous-gen pokemon can't be traded to BDSP yet.

Lusamine:
> That gen 3 Glaceon is legal because it could have been evolved in Sword/Shield with a stone. If you put a gen 7 exclusive move like Frost Breath it is flagged as illegal evolution because it should have leveled up to being able to evolve in USUM to learn that move.

The Glaceon I included is from gen 4, and it is in gen 7, where stone evolution does not exist.  The flag is incorrect.

image

It is captured as Eevee from Trophy Garden at level 22, leveled up once to 23 to become Glaceon, and then transferred. The transfer process sets the Met Level to 23.

You can also grab an Eevee from the encounter database and see that PKHeX also simulates this level behavior when transferred.

javierhimura:
You are right about Glaceon, it is the same edge case as Magnezone, it can evolve in gen 4 and then transfer to gen 5 or evolve in gen 5-7 with this one requiring to level up after the transfer. I checked Glaceon on a Gen 8 save instead of Gen 7 by mistake. I will continue debugging the code to fix this edge case too. I know these cases weren't detected as illegal before in PkHex, but after all the purpose of this change is to detect those things, so any edge cases found that don't work as they should I will look into it.

About BDSP Glaceon I agree with you. If that happens in the future this framework should be updated to include that case, but probably is best to wait to see how Nintendo implements trade between BDSP and Home. If they allow trade between BDSP and SwordShield both ways it will fall all checks on the same checks used in SwordShield (except maybe some untraded pokemon from BDSP). If trade is not two ways it opens a different can of worms, having some checks being different in Gen8 based on what game the pokemon it is, and probably this evolution won't be the only check that could be different between PreGen 8 in BDSP and PreGen8 in SwordShield.

javierhimura:
My mistake, Frost Breath shouldn't be illegal for this Glaceon. It needs to level up to evolve on Gen5-7 but in this case, it was transferred already evolved from Gen 4 up to 7 or 8. The same reason why is detected in USUM as legal. It was captured at level 22, evolved to level 23, transferred to black and white making met level 23 and then to gen 6 and finally gen 7 where it can learn Frost Breath.

This pokemon could be illegal if it should have evolved in gens 5 to 7, which happens with an exclusive Eevee move from gens 5 to 7. I am still checking this edge case.

Lusamine:
It may take a while if the goal is to quash every edge case.

Here's another Shedinja edge case.  Sport Ball Shedinja from HGSS had to have been evolved in DPPt and transferred, but is currently legal with Bide which is a pre-evo only level-up move at 29 in Gen 6.

Not to say that this PR should handle it, but memory legality with moves would eventually need to draw on this logic as well. I also included a Shedinja with Bide in the non-OT memory.  There are some nasty edge cases when memory legality is involved, e.g. having 4 past-gen only moves and a memory of a level up move in gen 6.

By the way, there is something going on with gen 1 Eevee being loaded into SWSH from Encounter database.  This error doesn't occur on master. I included the Eevee that triggers this exception.
image

I've also included a gen 1 level 42 Vaporeon with Haze, Mist, and Acid Armor. I am not seeing any way for these moves to coexist.
- RGB: acid armor level 42, haze level 44
- Yellow: acid armor level 47, haze level 42
- GSC: acid armor level 47, haze level 42

I'll be away for New Years (aka leaving now) so I figured I'd leave you these to think on.

[tests4.zip](https://github.com/kwsch/PKHeX/files/7796532/tests4.zip)

javierhimura:
Well, if the edge cases that are left are false negatives (things that are illegal but pkhex doesn't flag it as illegal and also didn't do it before the PR) I suppose it could be merged into Master and continue improving the checks one more edge cases are detected.

But other different things are legal things that are legal and this PR incorrectly flagged as illegal, and even worse those that break the analysts, the PR shouldn't be merged until being fixed.

The incompatible moves from Gen 1 and 2 are not handled by this PR because it involves more things than evolutions. I already thought about how to do that but it will be another PR after this one is completed. It requires all gen 1 and 2 moves to compare against each other to check if there is an incompatible pair and to check if two moves are incompatible get a list of all possible sources of the move (including level, evolution phase, game, type of learning). Even with a pokemon whose all 4 moves are from gen 1-2, it is only 12 pairs to compare. And probably some moves could be ignored because they are compatible with every other move (like most TM moves).

But that will happen next year.

javierhimura:
I have fixed all edge cases detected.

About move memories. It could be handled in this framework but I won't do it on this PR. I am not very familiar with pokemon memories. I have not played yet any pokemon switch games.

There was already on the encounter iteration moves a small check to determine if the pokemon can know the move it has on the memories about using a move and I moved inside the generation evolutions iteration.

The learn move memory check is not yet implemented but once done we could make the same as to know move memory and have both a complete memory check on the memory verifier and the small boolean check for move iteration. I don't make it right know because I am not entirely sure what could as a learned move.

The part about checking if the pokemon has learned memory without any gen 6 moves learned is actually the easiest part because the move analyzer already has the data of the maximum generation in which a move can be learned.

javierhimura:
I added the last change to show on the verbose report the generations in which the pokemon has evolved. In case someone makes a pokemon with exclusive pre-evolution moves it would be easier to see that it is illegal for having evolved one or more generations before being able to learn those moves.

All the false positives and errors detected are fixed and I can't think of more cases that should be illegal that aren't flagged, so the PR is ready to review for a merge.

Lusamine:
This is still flagging a lot of my Pokémon incorrectly.

From my tests3.zip, Glaceon with Frost Breath is again incorrectly flagged illegal in gen 8.
From my tests4.zip, past gen Shedinja with Bide is still incorrectly legal.

There are also a bunch of new false flags for Pokemon in gen 8 that do not evolve.
image

More legal files that are newly being flagged.
image

I enclosed 3 examples that are flagged incorrectly:
- Gen 7 HA Alolan Raichu in Gen 8 with 4 Gen 7 TM moves.
- An egg Ferroseed from Gen 7, which is legal in gen 7 and illegal in gen 8.
- A Vikavolt from Gen 7, which is legal in gen 7 and illegal in gen 8.

[tests5.zip](https://github.com/kwsch/PKHeX/files/7799967/tests5.zip)

kwsch:
With all these recurring edge cases, I'm pretty sure this isn't the right approach to flag things that aren't currently being flagged.

PKHeX currently lets these by as the algorithmic requirements to verify them are too muddy to be implemented, or too much work to accurately implement (as opposed to working on something else). Having hardcoded edge cases to account for every idiosyncrasy that may exist is extremely brittle and hard to maintain if new edge cases are discovered.

I tried improving on the evolution tree deficiency 2 years ago with the evotree branch, but it doesn't work well since there are so many dead-end cases for an evolution line to be legal. Looking at `master`'s current usages of all the evolution chains across generations, we have the following roadblocks to account for when trying to retrace an entity's progression from origin to current state:

* Transferring 5->6 setting an ability index value incompatible with other evolutions
* Evolving updating ability back to PID bit (CXD/Gen3 added abilities evolving in Gen4/5)
* Having certain evolution criteria (Beauty for Feebas) with evolution methods not available
* Learning moves in any generation (evolution requirement)
* Learning moves in any generation (current moves)
* Learning moves in specific generations (memories)
* Knowing moves in a specific generation or prior (memories)
* Transferring VC->7 with specific base friendship values
* Spawning a Shedinja in a specific game (Sport Ball HGSS, Affixed Ribbon/Mark SWSH)
* Having a evolution-variable-requirement ribbon which requires a specific evolution (Footprint in BDSP)

The correct way of solving this overall issue is instead crawling backwards from current->origin, to see if there is a path that exists where all of the above conditions are satisfied. Decreasing level, de-evolving, and transferring between games and generations are "moves" that a searcher would perform, and if there are no paths that satisfy all the above, then the algorithm gives up and indicates which roadblocks were unsatisfied.

That algorithm would be a pretty complex undertaking, and would replace the current move verification as well as the evolution verification, and nothing else. It would solve the Gen1/2 evolution exclusive move issues as well. However, similar algorithms have been implemented (namely, the Colo/XD shadow lock checks), so it's not impossible.
 :D

Since PKHeX has been super-stable with legality checking, we don't want to be introducing regressions (as mentioned in above comments). In the years we've spent automating legality checking, hardcoded approaches have been troublesome, whereas emulating the games and replicating human steps to validate has been easier to understand/implement/maintain. Feel free to discuss with us on Discord, we've been quite familiar with evolution edge cases for a while. With Pokémon Legends: Arceus coming in a few weeks, we don't want to merge this. What you've come up with is definitely great in exploring the toughness of this issue, there are now have a ton of edge cases documented whenever a devolving-descent algorithm is attempted 😄

javierhimura:
If I am being sincere when I started to make this I thought the only incompatibilities between generations were just the moves and that is. It has been after starting to test it with the help of Lusamine to see that there were many edge cases I didn't realize in the first place. I didn't even know about the ability patch because believe it or not I have not started playing gen 8 games until yesterday (literally). So I agree it has become a snowball that has grown too much and can't be merged just know.

PR #3379, opened 2022-01-31
Fix swarm seed location for pokemon platinum

edo9300:
The seed was being read/written to the wrong location of the save. I've tested this on my pokemon platinum savefile, diamond/pearl locations might be wrong as well, but I haven't checked those.


kwsch:
Original source for old (incorrect offsets):
https://projectpokemon.org/home/docs/gen-4/dp-save-structure-r74/
https://projectpokemon.org/home/docs/gen-4/platinum-save-structure-r81/

DP is likely wrong in PKHeX's current implementation as well.

kwsch:
DP offsets: 0x53C4 (safari) and 0x53C8 (swarm)
Pt offsets: 0x5660 (safari) and 0x5664 (swarm)

PR #3466, opened 2022-03-17
Fix BDSP Gen8Version Name Typo

foohyfooh:
Fix a typo in the Gen8Version enum's name unless this was meant to be a gem pun.


kwsch:
Nah, it's intended (gem as in brilliant diamond, shining pearl) similar to gems, especially with underground treasures. We didn't wanna call it "DprVersion"

PR #3506, opened 2022-05-31
HOME 2.0.0: Handle conversion behavior & restrictions

kwsch:
* Revises legality checks to account for traveling between the three game islands (PLA/BDSP/SWSH)
* Adds conversion mechanisms between the three formats, as well as flexible conversion options to backfill missing data (thanks GameFreak/ILCA for opting for lossy conversion instead of updating the games).
* Adds API abstractions for HOME data storage format (EKH/PKH format 1, aka EH1/PH1).
* Revises some APIs for better usage:
  - `PKM` now exposes a `Context` to indicate the isolation context for legality purposes.
  - Some method signatures have changed to accept `Context` or `GameVersion` instead of a vague `int` for Generation.
  - Evolution History is now tracked in the Legality parse for specific contexts, rather than only per generation.


Hooon42:
Is that mean we can transfer any PK7 and low to PB8 and PA8?

kwsch:
> Is that mean we can transfer any PK7 and low to PB8 and PA8?

Correct. The program will be up-to-date with current mechanics & limitations:
> Adds conversion mechanisms between the three formats

PR #3534, opened 2022-06-19
Fixed Layout of IV's to align with Default layout for Roaming Gen 4

ShadyRelapse-PP:


sora10pls:
Just so you know: internally, IVs and EVs have been ordered HP, Atk, Def, Spe, SpA, and SpD.

This is because prior to Gen 2, there was one unified "Special" stat, it came after Speed, so the ordering was retained and SpA/SpD were put after Spe. This has been the case since (as you can tell across a lot of instances in the PKHeX source code, and data structures in things like pk3DS/pkNX), and does not need revision.

ShadyRelapse-PP:
Only thing is iv's aren't listed in the Block Editor in the same way that they are in the Stats for editing PKMN.
example: here is the iv's of a roaming pokemon in the order of the Block Editor
19, 30, 14, 30, 1, 30
and here is the layout from RNG Reporter
19, 30, 14, 1, 30, 30

sora10pls:
Yes, the PKM editor shows it in the order in which the game does. However, Speed still comes before Special Attack. Your second proposed PR #3535 labels these IVs inaccurately. These changes are not needed.

PR #3536, opened 2022-06-20
BDSP: don't unlock National dex nor set Game Clear flag when unlocking events

ghost:
Azure Flute can be obtained before endgame (I just got mine, I'm at Floaroma Town), and the only requirement for obtaining Oak's Letter and Member Card was to unlock Mystery Gift, which you could do as soon as you had one badge. I think it would be better/safer to leave those alone, and let the game set them on its own as the player progresses instead.


Lusamine:
The idea is not to create a slippery slope where more and more flags must be changed to create a "sane" save file. The original implementation is expected to break story progress and has been viable since the game was released.  It assumes that you can physically go to where it is, and only need the flags to move the NPCs into the correct position.

PR #3549, opened 2022-07-12
Update .NET Framework 4.6 refs to 4.8

codemonkey85:
This PR is just to update the targets for .NET Framework 4.6 to target 4.8. I tested by building, running, and of course passing the unit tests.

I suppose you're trying to keep as wide a compatibility as possible, hence why .NET Framework is still supported, but is there any reason not to at least go to 4.8?

### DIFF
--- a/PKHeX.Core/PKHeX.Core.csproj
+++ b/PKHeX.Core/PKHeX.Core.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net46</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net48</TargetFrameworks>
     <LangVersion>10</LangVersion>
     <Nullable>enable</Nullable>
   </PropertyGroup>
--- a/PKHeX.Core/Saves/Encryption/SwishCrypto/SwishCrypto.cs
+++ b/PKHeX.Core/Saves/Encryption/SwishCrypto/SwishCrypto.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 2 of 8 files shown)


kwsch:
net46 is the highest it can be for running on Windows Vista! PKHeX used to be net40, but I dropped Windows XP support at the start of 2017 (so I could start using `Array.Empty<T>()` and other QoL improvements). XP had around 5% market share... nowadays, vista+XP is <0.5%, so not much of a loss to go higher.

I haven't felt the need to update to net47/net48 as any improvements to the BCL seemed unimportant. [Per the changelog](https://docs.microsoft.com/en-us/dotnet/framework/whats-new/), the only things higher net4* versions would give is better High DPI support, and removing concerns about FIPS mode. There's quite a few projects using PKHeX.Core, and they'll likely have to bump their required version too.

net4* works by picking out which runtime to use: config (not used) -> compiled target (if present) -> latest installed (if compiled not present). So, running PKHeX.WinForms on a PC with only net48... will work, and will run with net48.

There are still some users ([2 weeks ago!](https://projectpokemon.org/home/forums/topic/61624-pkhex-doesnt-work-for-some-reason/?do=findComment&comment=277540)) still without .NET 4.6+ installed, so I expect a bump to 4.8 would cause another round of people having issues (and new mono/wine setups needing to be made). If I disturb the process of laggards / non-Windows users, I'd rather do it only once... vvv

The current plan is to drop .NET Framework entirely in favor of .NET 7 when the program is updated for S/V support (4 months!). This will allow use of the new .NET runtime, unlocking new features & APIs (like default interface methods, and better performing span methods instead of array based). At the expense of those

kwsch:
tl;dr -- not now, but a jump is planned 4 months from now ;)

codemonkey85:
> When that happens, I'll just nag the non-Windows people to use a VM.

I mean... you could always rewrite the UI in .NET MAUI... :-P

PR #3575, opened 2022-08-27
Refactoring: Narrow some value types (Species, Move, Form)

kwsch:
When storing data for an encounter template, everything has a Species and Form. Currently, these two fields were saved as `Int32`'s. Recent refactorings for Evolutions developed the `UInt16`-Species, `UInt8`-Form minimalist size structure, and we can use the same narrowing to unify the codebase for these fields that are commonly stored in other objects.

Why the upheaval? When the program starts up, the runtime allocates memory for each new object that is created. Using less-wide types results in smaller objects, and less allocation (memory hungry application). Since PKHeX is downloaded over 100,000 times per month (and run via SysBots on Twitch and Discord), every little bit can help. Plus, doing things consistently across the codebase is less annoying :)

In the downtime between games, inefficiencies and old logic can be cleaned up. The original reason for int32-wide types for most common properties was because I initially had issues getting the `ComboBox` display logic to work with anything besides `int`'s; later became obvious that it was because `ComboItem` is of `{int,String}` type -- the `DisplayValue` type matters when it auto assigns. Lately we can just work around this behavior without having specialized `ComboItem` objects for wrapping different primitives like `ushort`.

In this pull request I've changed a ton of method signatures to reflect the more-narrow types of Species, Move# and Form; additionally, I've narrowed other large collections that stored lists of species / permitted values, and reworked them to be more performant with the latest API spaghetti that PKHeX provides. Roamer met locations, usually in a range of [max-min]<64, can be quickly checked using a bitflag operation on a `UInt64`. Other collections (like "Is this from Colosseum or XD") were eliminated -- shadow state is not transferred COLO<->XD, so having a Shadow ID or matching the met location from a gift/wild encounter is a sufficient check for "originated in XD".

Future updates may revise Met/Egg_Location, Gender, and Trainer ID, as well as splitting up static encounter arrays (no wasting time concatenating them).


Kermalis:
Hello, I noticed a few times that you have left these uint bound checks where ushort or byte is now used:
image

Just in case you didn't notice in the refactoring.

PR #3615, opened 2022-10-22
[Egg Encounters] Respect shiny/ability/IV criteria

Kermalis:


Kermalis:
Not sure what that "Fix FRLG flags" commit is doing here but I wanted to make this PR anyway since it's very useful

kwsch:
As mentioned on Discord:
- current Egg API is too generalized, needs to be split up
- current encounter yielding API needs to be more specific with less large arrays (berichan's use case for web)
- shouldn't use mutation/checking APIs in a generator API (separation of logic, hidden complexity)

ALM's implementation of post-creation mutations is an already sufficient implementation, which can adapt things further to the criteria. For now, it's best to rely on that and wait for refactorings when the time comes. Taking on that burden now (when ALM already exists) is not worth at the moment.

PR #3619, opened 2022-10-27
Fix: Check if property can be written to

LostQuasar:
Check if the property can be written to before attempting to write it.
This is important as PK1 does not define a set method for the Valid property and so trying to write that property causes the PropertyInfo.SetValue to throw `System.ArgumentException: 'Property set method not found.'`

### DIFF
--- a/PKHeX.Core/Editing/Bulk/BatchEditing.cs
+++ b/PKHeX.Core/Editing/Bulk/BatchEditing.cs
@@ -135,7 +135,7 @@ public static class BatchEditing
         }
         var props = Props[index];
         bool exits = props.TryGetValue(name, out pi);
-        if (!pi.CanWrite)
+        if (exits & !pi.CanWrite)
         {
             return false;
         }
--- a/PKHeX.Core/Editing/Bulk/BatchEditing.cs
+++ b/PKHeX.Core/Editing/Bulk/BatchEditing.cs
@@ -134,7 +134,12 @@ public static class BatchEditing
             return false;
         }
         var props = Props[index];
-        return props.TryGetValue(name, out pi);
+        bool exits = props.TryGetValue(name, out pi);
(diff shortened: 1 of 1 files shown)


LostQuasar:
https://github.com/kwsch/PKHeX/blob/master/PKHeX.Core/PKM/PK1.cs#L12

LostQuasar:
Another solution would be to introduce a set method for PK1 but I felt this was a more appropriate solution although it may be worth throwing a debug print in there to output that it was unable to write that value.

PKHeX.Core/Editing/Bulk/BatchEditing.cs
```diff
@@ -134,7 +134,12 @@ public static bool TryGetHasProperty(Type type, string name, [NotNullWhen(true)]
             return false;
         }
         var props = Props[index];
-        return props.TryGetValue(name, out pi);
+        bool exits = props.TryGetValue(name, out pi);
+        if (!pi.CanWrite)
```
LostQuasar, line 138:
```suggestion
        if (exits & !pi.CanWrite)
```

kwsch:
This breaks the upstream use of `TryGetHasProperty` -- it is used to display what the current value of a property is in the GUI.

What is your use case of this? There are no usages of this method that can result in uncertain `CanWrite` status.

LostQuasar:
Using this for EntityConverter.ConvertToType which uses TransferPropertiesWithReflection

kwsch:
Looks like it's because `TransferPropertiesWithReflection` looks at the entire type hierarchy, which has a setter via `PKM`. I will add the check there.

LostQuasar:
Cool, ty. Pretty niche use case but basically what's causing this error is converting backward into gen 1 which is the only time this is actually an issue.

PR #3634, opened 2022-11-25
Minor Temp Tests Update in Legality from BDSP to SV and added vscode directory in gitignore

RaigaX9:
Minor Temp Tests Update in Legality from BDSP to SV and added vscode directory in gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -252,3 +252,4 @@ pip-log.txt
 #################
 
 *.userprefs
+/.vscode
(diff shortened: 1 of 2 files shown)


kwsch:
Why?

The old test ensures that BD/SP egg moves that are (impossible to obtain) are not recognized as valid by the program.

The changes you've made are just testing if specific non-egg-moves are recognized as valid, but the comment about HOME is incorrect. The encounter generator will successfully generate those starting moves too, because Sprigatito can learn Leafage at any level, so the test will fail.

PR #3707, opened 2022-12-30
Update lang_zh.txt

kckxpyr1:


Lusamine:
The English button only says "Raids", which I believe should be 团战?
image
It looks like your translation adds "Editor" to the end.

FireJoker:
I think "团体战" could be better.

Kitcham:
I think "團體戰" is better for CHT translation, and "团体战" is better for CHS translation.

Both translation texts are taken from the [official Pokémon website in Hong Kong](https://sv-news.pokemon.co.jp/tc/page/21.html). "太晶" means "Tera" in English, "團體戰" and "团体战" mean "Raid".

If possible, can you please update the traditional Chinese version as well? @kckxpyr1  Thanks!

CHT
CHS

Lusamine:
This was handled in https://github.com/kwsch/PKHeX/pull/3717 so I'm closing this. Thanks to everyone for the input!

PR #3736, opened 2023-01-25
Remove the useless blank line in text_Natures_zh2.txt

Kitcham:
Remove the useless blank line to fix the Natures drop-down selection menu issue, which may be introduced from the last modification.

- Traditional Chinese language in now version:
In Traditional Chinese language

- Other languages in now version:
In other languages


jimmyorz:
I'm confused about this too, I only edited the Relaxed line.  but the last line always shows that there has been a change

Kitcham:
> I'm confused about this too, I only edited the Relaxed line. but the last line always shows that there has been a change

Oh, maybe you used the GitHub online editor. If you use the online editor instead of some local text editors, some problems may arise.

You can try to use some professional text editors, like VS Code, to open the text file. Then you will see that there is a blank line at the end of the text.

blank line

Kitcham:
And some translation tweaks for better display because the text could not display correctly before.
tweaks

PR #3751, opened 2023-01-30
Update text_ItemsG3_ja.txt

mi-ya1987:


mi-ya1987:
i making gen3's jp script file , slowly.

PR #3779, opened 2023-02-15
Update flags_e_en.txt

mi-ya1987:


Lusamine:
The item was named "Thunderstone" back then, and the current English text is "Thunder Stone" with both words capitalized. If the goal is to standardize to the current text (PKHeX shows "Thunder Stone" even in Emerald, and the flag right below it uses "Fire Stone"), we should probably use the latter formatting.

mi-ya1987:
I first made it by quoting flags_frlg_en.txt file.
Then I used the find and replace function.

In flags_frlg_en.txt it was described as Thunder Stone.
That's why I couldn't replace it, and I thought that emerald was wrong.

mi-ya1987:
en

Lusamine:
image
What I'm saying is this should be "Thunder Stone" like in your picture, not "Thunder stone".

Lusamine:
PKHeX uses the most recent names even for older games.
You can check by setting the save file to an older game version and trying to add a Thunder Stone.
image

mi-ya1987:
It's an English version problem.
en4
If you said "PKHeX uses the modern name for an old game" , you should have corrected it to "Thunder Stone" instead of reverting.
There are still some variations in the English notation.
en5
This is not a problem in the Japanese version, but it causes a problem when making the Japanese version.
In Japanese, all generations are "かみなりのいし".

Lusamine:
It would make more sense to correct it to a proper format if you were going to correct it.

By the way, there should be a way for you to edit multiple files and open one pull request instead of opening all of these pull requests.  You would fork the repo and edit on your version first, then open a single pull request with all changes.

mi-ya1987:
Thanks for the video.

If you say, "It would make more sense to correct it to a proper format if you were going to correct it." so why don't you correct it?
A reasonable and appropriate correct is not to undo the wrong correct, but to correct it right away.
I corrected another mistake at the same time and it never corrected.
en6
I made a mistake in correcting it, but I could have show you where it went wrong.
Why didn't you correct even one place?
If the correction is incorrect, it should be corrected rather than simply rejected.
If you could change "Thunderstone" to "Thunder Stone" and "Elixier" to "Elixir", we'd be fine.
It would have been nice if you could have pointed out that it was "Thunder Stone" instead of "Thunder stone" after that correct.
Otherwise, the issue will remain unresolved.
If you say it's not your job, I have to say it's not mine either.
What I want to do is translate into Japanese, not point out spelling mistakes in English.

I understand that you pointed out the mistake, I appreciate that.
However, I don't understand why you left the mistake and didn't correct it.
Are you in a position where your only role is to point out mistakes and you are forbidden to correct them?
If so, I convince.

Lusamine:
This is your pull request so I left it up to you to correct it how you want.
The flag files are very old and many of them are edited by other people who did not follow a standard or proofread their edits.

You are going through English phrases in a Japanese file and making them less correct.
Both "Thunderstone" and "Thunder Stone" would be more correct than "Thunder stone."

I do not spend my time going through the flag files to proofread them.

I am recommending that you either revert it to "Thunderstone" if you do not care about English, or change it to the current term which is "Thunder Stone."

You can pick which one you prefer to do.

mi-ya1987:
How about "Elixier" to "Elixir"?
When you realize this mistake, do you not correct it yourself?

Lusamine:
I did not look at this file until you tried to fix it here.
This is your pull request, so you can fix it if you want.
If you don't know how to fix it, then I can fix it for you.

Lusamine:
I pushed a commit that corrects the spelling to the current name. After the other pull requests are closed, you will probably be better able to edit your pull requests by working on your repo instead of opening many pull requests.

Lusamine:
I see that you opened a separate one here: 
https://github.com/kwsch/PKHeX/pull/3787

Normally, you would correct the first pull request or close it first, not open two at the same time changing the same thing. I am going to close this one and merge the other one.

PR #3810, opened 2023-02-25
Add Geonet location editing for Gen 4/5

abcboy101:
- Added getters/setters for the player's registered Geonet country/region to SAV4/SAV5
- Added a country/region select to SAV_SimpleTrainer for Gen 4/5 saves
  - Formatted the text resources to try and match locale3DS
  - I'm not that familiar with the form designer so it's possible that I might have messed something up while editing the form
- Noticed a few 3DS sub-regions were missing, so added them as well


kwsch:
Nice!

Since commas are now present in some of the subregions, I'll do a follow-up commit to redo all the txt files to be tab-separated instead of comma-separated.

PR #3812, opened 2023-02-26
Update lang_zh.txt

kckxpyr1:
This is more in line with the Chinese translation, I hope it can be modified


Lusamine:
This was discussed previously as the better option with official images.

https://github.com/kwsch/PKHeX/pull/3707#issuecomment-1368883329

Someone else changed it before you did so I closed the last pull request. Is there a reason this translation is better than the one in the link?  I do not want to have multiple Chinese users trying to change the same text back and forth.

kckxpyr1:
Although the "raids" in English is the meaning of "团体战", the actual use is a tool for editing team war content, and it is not appropriate to simply translate as "团体战". I am based on the actual content and use of use The effect was considered to join the translation of "Editor", so I would think that the modification was more consistent. I hope that the author can greatly understand

kckxpyr1:
Although the "raids" in English is the meaning of "团体战", the actual use is a tool for editing team war content, and it is not appropriate to simply translate as "团体战". I am based on the actual content and use of use The effect was considered to join the translation of "Editor", so I would think that the modification was more consistent. I was a bit busy during the time before and did not pass the modification or get a reply, so I didn’t notice it. I don't know if someone else made an opinion in the revision before me, so I did not come forward to reply to the comment. I was embarrassed. I hope that the author can understand greatly

Lusamine:
None of the other buttons on that page use "Editor" and it would be redundant to add it to all of them. It should be understood that the different buttons are all editors. I am leaving the translation as it is to stay consistent with the other languages.

PR #3815, opened 2023-02-27
Fix Jiseok Garganacl start date

Manu098vm:


sora10pls:
Zoroark end date is already accounted for in the sv120 branch. It should be 3/1/2024 due to end date tolerance (time zones).

Manu098vm:
Aighty, I adjusted the end date. Feel free to discard this PR eventually, if you feel more comfortable to edit the code by yourself.
The start date for the Garganacl have to be adjusted as per my commit tho. Other than the year which was wrong, due to timezones the event was already available on the 16th in some places. Where I live for example, the code was announced on the 16th around 22pm.

Lusamine:
I would suggest only correcting the Garganacl on this PR since the Zoroark is already correctly handled on the testing branch that will be merged.

PR #3840, opened 2023-03-19
Filter Block Dump data

javierhimura:
I noticed after the pull adding Vivillon Blocks that the Dropdown with the save blocks has no filter option.

I added a filter for it. If you write a text and press enter, it will filter the items that contain that text.
I kept the original functionality of selecting an item if the input text is a hexadecimal string.

So, now if you type, for example, "Vivillon" and press enter, it will filter the blocks related to Vivillon GO connection.


Lusamine:
From what I can tell, this doesn't actually do anything?
It also broke the previous block key search.
At some point, it popped up an exception about a key not found but I am not sure how to replicate that.

block filter

javierhimura:
I have noticed the problem with the previous block key search, some hex values are converted to negative numbers, so I changed it to check is not 0.

Also fixed the text filter. After making and testing the change, I mistakenly inverted the comparison while trying to clean the code.

Lusamine:
There's some misbehavior if you filter it once, then try to use it for another search filter. Seems that it doesn't clear the filter unless you do it a few times, or you hit enter on a completely blank box.

refilter
Here it works fine for "Vivillon" the first time, then ignores the block key, and by the end it no longer allows me to search "Vivillon".

Lusamine:
Seems like the commit https://github.com/kwsch/PKHeX/pull/3840/commits/6e97ee2c90618f924ada9443e93a0713e6d4c2e6 fixed one of the issues I just brought up; I tested before you had pushed it.

I notice the search term is ignored if you have the dropdown open when searching.

javierhimura:
I was able to replicate that last issue but the problem is that when you write in the search term with the dropdown open and press enter the text box is received empty in the event. That also happens with the master branch if you write a hex value with the dropdown opened.

PR #3846, opened 2023-03-20
Fix first character of Japanese country/region names getting cut off

abcboy101:
When the interface language is set to Japanese, it incorrectly displayed `日本` or `アメリカ` as just `本` or `メリカ` in both the Trainer Data editor and the PK6/7 editor.


kwsch:
I decided to rewrite the method to be a little less obtuse; the sneaky `+ 1` shifting inside the original method was not obvious :)

PR #3867, opened 2023-04-23
Added ability to edit fly locations in Crystal

Adam887:
Added a Misc page for Gen 2 with the fly locations in Crystal. Tried copying the UI and logic from Gen 4

### DIFF
--- a/PKHeX.Core/Saves/SAV2.cs
+++ b/PKHeX.Core/Saves/SAV2.cs
@@ -715,36 +715,6 @@ public sealed class SAV2 : SaveFile, ILangDeviantSave, IEventFlagArray, IEventWo
         SetFlag(EventFlag + (flagNumber >> 3), flagNumber & 7, value);
     }
 
-    public static uint[] GetCrystalFlyFlags()
-    {
-        return new uint[] {
-            0x1,        //Lake of Rage
-            0x2,        //Blackthorn City
-            0x4,        //Silver Cave
-            0x100,      //Violet City
-            0x400,      //Azalea Town
-            0x800,      //Cianwood City
-            0x1000,      //Goldenrod City
-            0x2000,     //Olivine City
-            0x4000,     //Ecruteak City
-            0x8000,     //Mahogony Town
-            0x10000,    //Lavender Town
-            0x20000,    //Saffron City
-            0x40000,    //Celadon City
-            0x80000,    //Fuchsia City
-            0x100000,   //Cinnabar Island
-            0x200000,   //Indigo Plateau
-            0x400000,   //New Bark Town
-            0x800000,   //Cherrygrove City
-            0x4000000,  //Pallet Town
-            0x8000000,  //Viridian City
-            0x10000000,  //Pewter City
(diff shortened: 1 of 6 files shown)


kwsch:
Hey, sorry for the lack of feedback!

I have a few issues with the current implementation:

* Only supports International (not-JPN) Crystal and all other Gen2 savefiles that try to open the editor will crash (because the offset is -1)
* There are more fly-flags than those you've handled, and I think the flags are stored in little-endian order (not [big-endian](https://github.com/kwsch/PKHeX/pull/3867/files#diff-1608232fe16b1f37f7de2a35781332eb422137edc6e25ebc2c9db655a2510ff9R96-R120) as you have them mapped):
1. [Flypoint constants](https://github.com/pret/pokecrystal/blob/354fefaefbba77a3408fa9edb5cc4393b9c702ab/constants/map_data_constants.asm#L109-L139)
2. [Flypoint flag ordering & count](https://github.com/pret/pokecrystal/blob/354fefaefbba77a3408fa9edb5cc4393b9c702ab/constants/map_data_constants.asm#L73-L105)
3. [Savedata flag storage](https://github.com/pret/pokecrystal/blob/354fefaefbba77a3408fa9edb5cc4393b9c702ab/ram/wram.asm#L3276)

I'll take a more detailed look to see how support can be added for the other Gen2 games. Ideally, we end up with a get/set method in SAV2 for individual flypoint interactions, for all Gen2 games. Maybe also for Gen1.

Adam887:
> Hey, sorry for the lack of feedback!
> 
> I have a few issues with the current implementation:
> 
> * Only supports International (not-JPN) Crystal and all other Gen2 savefiles that try to open the editor will crash (because the offset is -1)
> * There are more fly-flags than those you've handled (Kanto!), and I think the flags are stored in little-endian order (not [big-endian](https://github.com/kwsch/PKHeX/pull/3867/files#diff-1608232fe16b1f37f7de2a35781332eb422137edc6e25ebc2c9db655a2510ff9R96-R120) as you have them mapped):
> 1. [Flypoint constants](https://github.com/pret/pokecrystal/blob/354fefaefbba77a3408fa9edb5cc4393b9c702ab/constants/map_data_constants.asm#L109-L139)
> 2. [Flypoint flag ordering & count](https://github.com/pret/pokecrystal/blob/354fefaefbba77a3408fa9edb5cc4393b9c702ab/constants/map_data_constants.asm#L73-L105)
> 3. [Savedata flag storage](https://github.com/pret/pokecrystal/blob/354fefaefbba77a3408fa9edb5cc4393b9c702ab/ram/wram.asm#L3276)
> 
> I'll take a more detailed look to see how support can be added for the other Gen2 games. Ideally, we end up with a get/set method in SAV2 for individual flypoint interactions, for all Gen2 games. Maybe also for Gen1.

Thanks for the response, I can make those changes adding support for the JPN version and adding in the missing flags. I can also look at adding support for the other gen 2 games. However, I'm also happy to leave this work if you or someone else can look into it more and finding a better way of doing it.

PR #3873, opened 2023-04-30
Added Boxes to view Current Trainer photo and Trainer Icons

pasqualenardiello:
- Added possibility to see Trainer Current Photo
- Added possibility to see Current and Initial Icons
- Added compact layout for the 3 images
- Added dynamic form resizing in relation to Current photo width
- Readded Trainer TID/SID boxes
catch2
catch3

### DIFF
--- a/PKHeX.Core/Saves/Access/SaveBlockAccessor9SV.cs
+++ b/PKHeX.Core/Saves/Access/SaveBlockAccessor9SV.cs
@@ -96,14 +96,14 @@ public sealed class SaveBlockAccessor9SV : SCBlockAccessor, ISaveBlock9Main
     private const uint KLastGreenPosition = 0x5C6F8291; // PlayerSave_LastGreenPos
 
     // Fashion
-    private const uint KFashionUnlockedEyewear = 0xCBA20ED5; // 1000-1999
-    private const uint KFashionUnlockedGloves = 0x581667B1; // 2000-2999
-    private const uint KFashionUnlockedBag = 0x5D6F8110; // 3000-3999
-    private const uint KFashionUnlockedFootwear = 0x0221A618; // 4000-4999
-    private const uint KFashionUnlockedHeadwear = 0x860CD8FB; // 5000-5999
-    private const uint KFashionUnlockedLegwear = 0xD186222E; // 6000-6999
-    private const uint KFashionUnlockedUniform = 0x78FF2CB2; // 7000-7999
-    private const uint KFashionUnlockedPhoneCase = 0xED0AC675; // 8000-8999
+    public const uint KFashionUnlockedEyewear = 0xCBA20ED5; // 1000-1999
+    public const uint KFashionUnlockedGloves = 0x581667B1; // 2000-2999
+    public const uint KFashionUnlockedBag = 0x5D6F8110; // 3000-3999
+    public const uint KFashionUnlockedFootwear = 0x0221A618; // 4000-4999
+    public const uint KFashionUnlockedHeadwear = 0x860CD8FB; // 5000-5999
+    public const uint KFashionUnlockedLegwear = 0xD186222E; // 6000-6999
+    public const uint KFashionUnlockedUniform = 0x78FF2CB2; // 7000-7999
+    public const uint KFashionUnlockedPhoneCase = 0xED0AC675; // 8000-8999
 
     // Profile Picture
     public const uint KPictureProfileCurrentWidth = 0xFEAA87DA;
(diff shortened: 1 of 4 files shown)


sora10pls:
The save blocks that you've hardcoded in this pull request are already labeled in PKHeX, so should probably grab from that:
https://github.com/kwsch/PKHeX/blob/7e2552d9186d38e730742a8ee81a37a699637a13/PKHeX.Core/Saves/Access/SaveBlockAccessor9SV.cs#L108-L128

In my opinion, the trainer icons should be moved to a separate tab of their own. I dislike the heavily increased size of the trainer info editor. Other than that, all looks good, and is functional on my end as well.

pasqualenardiello:
Got it, I'll make the modifications ASAP

pasqualenardiello:
I made the modifications requested and also added the possibility to save images when clicking on them. Furthermore I made a small workaround for uniforms. Since editing player gender caused a bug by which, even if you clicked on one of them in-game, you couldn't change your uniform because the game mismatched the uniforms between the two genders. Also, I noticed that uniforms gender ids are the same for both Scarlet/Violet, so the only thing to do is inject the correct set and the game automatically change palettes. I also put a conditional to check if the new set of uniforms is present and update it only if it's the case.

sora10pls:
Not sure if the uniform inclusion belongs in this pull request. PKHeX generally disallows all gender swapping as of Gen 6 due to there being a number of issues with it compared to past games where it was just a simple 2D sprite.

Your current implementation checks if unlocked uniform slot 5 is not 0xFF (keep in mind that uniforms are stored as a value in the thousands, and each entry is ulong 8 bytes), but what happens when the first wave of DLC comes out and the player never claims their DLC purchase uniforms, but instead claims their Kitakami uniform? Your code would overwrite those uniforms and force them to reimport a save backup. I think the better solution is to disallow gender swapping like the other games as it requires more legwork to make it function as intended.

pasqualenardiello:
Ok, I'll revert the uniform changes then. Regarding the disallowing of gender swapping it is not implemented yet but I'll wait the end of this pull request to implement it (if someday else wouldn't already do it, of course).

pasqualenardiello:
Hi, don't want to bother you with other commits, but I have mapped all the clothing options in S/V and inserted the injector in the trainer Form. May I add the new edits in this pull request or shall I wait till its closing?

foohyfooh:
It would be wise to look into using branches because you have too different sets of features in this pull request. This would allow you to leave the profile picture changes alone and work on the fashion changes independently.

https://www.atlassian.com/git/tutorials/using-branches
https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow

Make a copy of your master branch with`git checkout -b master_backup` since we are going to do some destructive changes

Make a new copy for the profile picture changes with `git checkout -b sv_profile_picture`
Do a rebase to remove the fashion related changes with `git rebase -i 8f09fec1ec6538236cc8f4c13c27fc1ce56019ca`
In the editor put the following i.e. change the pick to drop on each line
```
drop 0aedf3899 Fixed uniforms bug when changing player gender
drop 2abfe29a1 Reverted uniform logic until gender swapping is disallowef
drop 3a8014d55 Revamped Fashion for Gen9 with Injector and locked Trainer Gender Swap
```
Verify that everything still works

Go back to master with `git checkout master` and make a new copy for the fashion changes with `git checkout -b sv_fashion_injector`
Do a rebase to remove the fashion changes with `git rebase -X ours -i 7e2552d9186d38e730742a8ee81a37a699637a13`
In the editor put the following i.e. change the first two from pick to drop
```
drop f2d98ede2 Added Boxes to view Current Trainer photo and Trainer Icons
drop 8f09fec1e Adjusted Images Visualization
pick 0aedf3899 Fixed uniforms bug when changing player gender
pick 2abfe29a1 Reverted uniform logic until gender swapping is disallowef
pick 3a8014d55 Revamped Fashion for Gen9 with Injector and locked Trainer Gender Swap
```
Verify that everything still works

Now comes the destructive parts
While on the sv_fashion_injector branch, delete master with `git branch -D master`
Recreate master using sv_profile_picture with `git checkout -b master sv_profile_picture`
Force push master to handle the current open PR with `git push -f`

Now you can work on sv_fashion_injector while you wait for the current PR to be reviewed
and to have it on GitHub use `git push -u origin sv_fashion_injector`

pasqualenardiello:
ok, i'll close this pull request then, thanks for the response

foohyfooh:
The point of the destructive changes were so you didn't have to close this existing pull request.

PR #3880, opened 2023-05-12
Fix Greyscale Calculation

foohyfooh:
The calculation done appears to be both luminosity and average methods mixed together which is causing the image to be much darker but I am assuming it should only be luminosity.

### DIFF
--- a/PKHeX.Drawing/ImageUtil.cs
+++ b/PKHeX.Drawing/ImageUtil.cs
@@ -231,7 +231,7 @@ public static class ImageUtil
         {
             if (data[i + 3] == 0)
                 continue;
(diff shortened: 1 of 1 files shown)


kwsch:
image
image
image

The intent for the method is to hard-differentiate totems. Due to some forms (like Raticate or Marowak) not having much color to begin with, the idea is to be much more visually different than just grayscale. The 3 is essentially a fudge factor to make things dark enough to have a consistent "totem" appearance.

I think this "grayscale" design was before the C/XD shadow aura, so it might be better to just use an orange aura to redraw instead.

kwsch:
Revised appearance as described in previous comment:
image

PR #3900, opened 2023-06-04
add pichu wc9 dist. date

Bappsack:


Bappsack:
Serebii lied to me. Also added the bronzong

https://projectpokemon.org/home/files/file/5022-wc-1002-michina-pichu/
https://projectpokemon.org/home/files/file/5021-wc-1003-marcus-bronzong/

PR #3932, opened 2023-06-30
add Paul Chua's Shiny Arcanine server data

Joseph11024:


sora10pls:
June 29th not possible, July 5th not possible - no eligible time zones.

Issue #3890, opened 2023-05-26
International Crystal checksum validation typo corrupting save data


pikavees:
**Describe the bug**
In SAV2Offsets.cs, line 103, the OverallChecksumPosition2 offset for international Crystal is erroneously set to 0x7E6D instead of the expected 0x1F0D. Since PKHeX only checks the first stored checksum at 0x2D0D, this isn't caught by the "Verify Checksums" button. In niche situations, such as the first checksum getting corrupted somehow between exporting through PKHeX and loading it in-game, this can cause complete save data corruption.

**To Reproduce**
Since PKHeX only validates the first checksum, to actually see this behavior, some workarounds have to be done.
1. Dump an international Crystal save file (I used English)
2. Change the checksum at 0x1F0D to be some incorrect value with a hex editor (I used "0xCC 0xCC")
3. Open the edited save in PKHeX and make some sort of change that would affect the checksum, export the save
4. Open the exported save in a hex editor - 0x1F0D will not have changed and will still be set to the wrong value from before, even though PKHeX is supposed to change it
5. Go to 0x2D0D and manually change the checksum there to some other wrong value. (I used "0xCC 0xCC" again)
6. Load the save file in-game. It will say the save file is corrupted and force you to make a new one. If the other checksum at 0x1F0D was changed properly by PKHeX, this wouldn't happen, since it would use the second set of data if 0x1F0D were valid, even if 0x2D0D wasn't.

**Expected behavior**
PKHeX should be changing the second international Crystal checksum at 0x1F0D, not 0x7E6D. I also think it's a bit strange that the Verify Checksums button in the SAV tab only validates the first checksum, and not the second.

**Screenshots**
Here's proof of my exported save file being corrupted after I only changed the checksum at 0x2D0D. The checksum at 0x1F0D remained untouched by PKHeX and as such the save file was corrupted.
![image](https://github.com/kwsch/PKHeX/assets/134743246/5cfaeba2-1a08-4e91-a34d-ac5e48340851)

**Additional context**
I happened to notice this while rummaging through PKHeX's source code to try and understand the gen 2 checksum validation better. I would have submitted a Pull request but I'm not comfortable with using GitHub at this time and thought it more appropriate to submit a bug report instead.

kwsch:
Closed by #3939

PR #3939, opened 2023-07-08
Crystal (International) Checksum 2 Fix

Zazsona:
'Ello 'ello! Me again!

This PR addresses the issues @pikavees reported in #3890! - Looks like it was just a classic typo problem! I was able to recreate their scenario and confirmed that with these changes, it no longer occurs!
I've also added their suggestion to rejig the Checksum verification logic a bit to compare against both values, with it reporting which one(s) pass/fail!

Ta'!

### DIFF
--- a/PKHeX.Core/Saves/SAV2.cs
+++ b/PKHeX.Core/Saves/SAV2.cs
@@ -308,17 +308,22 @@ public sealed class SAV2 : SaveFile, ILangDeviantSave, IEventFlagArray, IEventWo
         WriteUInt16LittleEndian(Data.AsSpan(Offsets.OverallChecksumPosition2), accum);
     }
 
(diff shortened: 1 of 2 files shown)


kwsch:
Shouldn't the Korean/JPN secondary offsets be changed, as well as all for Gold/Silver?

```cs
        OverallChecksumPosition2 = 0x7E6D;
        OverallChecksumPosition2 = 0x7F0D;
        OverallChecksumPosition2 = 0x7E6B;
```

Initial support had yoinked the offsets/ranges from [Bulbapedia's Gen2 structure page](https://bulbapedia.bulbagarden.net/wiki/Save_data_structure_(Generation_II)#File_structure)... not sure how correct it is for this checksum/mirror?

PR #3944, opened 2023-07-23
Update gen 2 event flags and PLA blocks

Momonopopotee:
Hi, as the title says I'm adding new gen 2 events flags as well as other translations to include them. I also C&P'd many of the Spanish translations from Crystal into Gold/Silver's seeing as it was straightforward.

Additionally, I updated 2 block keys for PLA accordingly relating to the Origin Dialga/Palkia story events.

Have a nice day!

### DIFF
--- a/PKHeX.Core/Resources/text/script/gen2/flags_c_en.txt
+++ b/PKHeX.Core/Resources/text/script/gen2/flags_c_en.txt
@@ -78,6 +78,8 @@
 0084	s	Togepi Hatched
 0085	s	Showed Togepi to Elm
 0086	s	Received Everstone from Elm
+1193	t	Defeated Blackbelt Kiyo (Can receive Tyrogue)
+0097	g	Received Tyrogue
 0046	s	Unown appeared in Ruins of Alph
 0098	s	Met Frieda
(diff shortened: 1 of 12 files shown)


sora10pls:
Why rename the story related Origin Forme Dialga/Palkia block keys, but none of the Noble Pokémon? All of them are story related, and can be rematched in the post-game. I don’t believe specifying story is necessary.

Momonopopotee:
> Why rename the story related Origin Forme Dialga/Palkia block keys, but none of the Noble Pokémon? All of them are story related, and can be rematched in the post-game. I don’t believe specifying story is necessary.

Yeah, I thought that because some people who'd mess with them might think they need to enable both when it's choice determined via this block: *Bool KChoseDiamondClanLeader. They're also the final boss required to see the credits roll depending on said choice.

So should I add the "Story" part to the other Nobles or remove it altogether (leaving the notes for reference)?

sora10pls:
I would revert the name to what it originally was (the less words needed to explain, the better) but keep your additional comments.

Momonopopotee:
Done, thanks for letting me know :)

PR #3949, opened 2023-08-02
Add battle subway trophy flags (BW)

alvv-z:
I added the flags to the other translations in English so that someone more well versed in the respective language can add a proper translation.


sora10pls:
“posess” should be “possess”

alvv-z:
> “posess” should be “possess”

I apologize, I have adjusted the files accordingly.

PR #3951, opened 2023-08-03
Add Foongus and Amoongus flags （BW）

ppllouf:
Add Foongus and Amoongus flags

Fix translation errors

Translation New flags in CHS


sora10pls:
"Amoongus" should be "Amoonguss"

Issue #3941, opened 2023-07-11
Hayley's trades have 3 perfect IVs when obtained via the encounter database


Atrius97:
**Describe the problem**
Hayley's trades all appear to have 3 perfect IVs minimum when obtained from the encounter database. To my knowledge, these Pokémon weren't guaranteed any perfect IVs.

Example legitimate Mews from PPORG: https://projectpokemon.org/home/files/file/330-mew/

**Expected behavior**
Hayley's trades should be generated with random IVs, no minimum number of perfect IVs.

PR #3955, opened 2023-08-09
Encounter Templates: Searching and Creating

kwsch:
Another fun adventure into optimizing legality checks!

## Backstory

When PKHeX was initially adding support for the games that it supported, the implementations were very straightforward. Need to find a possible encounter the Pokémon originated from? Iterate through a list of every single encounter, and whenever we find a good enough match, do the rest of our checks. This worked fine when the program only supported two sets of games (X/Y, OR/AS) as the list of encounters was very small.

At the same time, abstractions were added to differentiate the types of encounters a player could obtain. These encounters were:
1. Wild Encounter (encounter slot)
2. Static Encounter (captures and gifts from NPCs)
3. Trade Encounter (NPC with fixed OT/Nickname)
4. Mystery Gifts (WC6)
5. Egg (daycare)

These abstractions have served as the base "encounter types" for many years, but their overuse as a base abstract class did not fit well as more games were added. Some games might have fixed Nicknames and OT names for NPC trades, while others may only be OT name. Static encounters have different PID/IV generating types, requiring branched checking code. After a few years, Auto Legality Mod (ALM) wanted to use PKHeX to legalize existing PKM, and the Encounter Template => PKM logic was created as a result. Rather than updating existing PKM, new PKM were created from the template with criteria from a Showdown Set.

## Implementation Issues

As previously alluded to, when searching for an encounter, PKHeX iterates through a list of all possible encounters. Over time, this iteration has been split so that it only iterates for the games it needs to check, but most arrays it looped through were a single "master array". Having a master array per encounter type is okay, but we can do better. Combining arrays is unnecessary allocation. If we used a master array of type `IEncounterable[]`, every time we check compatibility we'd have a virtual call, and if we use IEnumerable<T> we'll have heap allocations for our "find best match" search. So, how can we improve things?

## Enter: State Machines

[Finite-state machines](https://en.wikipedia.org/wiki/Finite-state_machine) are wonderful for modeling iteration. C#'s `yield` keyword causes the [compiler to generate a state machine of its own](https://learn.microsoft.com/en-us/archive/msdn-magazine/2017/june/essential-net-custom-iterators-with-yield), so it's not like PKHeX wasn't doing it already. But the compiler implementation creates it as a class, with each array iterated stored separately in a field.

What do we really need to keep track of to iterate all encounters?
1. What array we're iterating (enum switch case)
2. What index within that array (and sub-index for encounter slots)
3. A deferred encounter and its match rating (to keep track of a "bad" match in hope of a better one)
4. Cached metadata about the PKM (so we don't need to recalculate every time)

## Refactor -- tl;dr

We implement simple state machine iterators to iterate through every split type encounter array, and more finely control the path we iterate through. And, by using generics, we can have the compiler generate optimized code to avoid virtual calls.

In addition to this, we shift away from the big-5 encounter types listed above, and not inherit from an abstract class. This allows for creating a PK* of a specific type and directly writing properties (no virtual calls). Plus we can now fine-tune each encounter type to call specific code, and not have to worry about future game encounter types bothering the generation routines.

Future re-works can be done:
* GetPossible/GetEncounter(GetBestMatch) don't heap allocate the IEnumerator, and do the iteration/finding within the method (zero/minimal allocation!)
* GetMatchRating can return a wider range of enum values, so we can use the enum value as indication to the user why the encounter doesn't fully match what is needed (eg gender locked encounter needs the other gender).
* Combining the Gen8b/Gen9 SW/SH-specific encounter iterators back into the mainline iterators (like Gen8a).


kwsch:
Resolves https://github.com/kwsch/PKHeX/issues/3941

PR #3959, opened 2023-08-13
Add Eduardo Gastrodon eligible date

lGodHatesMel:


sora10pls:
Please don’t copy/paste reasoning for past entries if the data you input does not follow that reasoning.

8/10/2023 is the earliest possible start date, not because of GMT-10 regions.
8/15/2023 is the correct end date.

PR #3983, opened 2023-09-09
Add Crystal Mystery Gift item in Simple Editor

RainingChain:
For Pokemon Crystal internation release, add the option to edit the Mystery Gift item in the Simple Editor.


Bl4ckSh4rk:
For Japanese Crystal the offset is 0xB52.

PR #4035, opened 2023-10-13
Add Trixie Mimikyu Eligible Date Range

BakaKaito:
Also Update wc9.pkl


sora10pls:
Wrong end date; codes are redeemable until December 31, 2023 @ 23:59 UTC.

BakaKaito:
Fixed. That's my bad for going off Serebii's End date and not considering the serial code's expiration 😅

sora10pls:
Serebii's date range is the only set of dates that were announced (distribution window), which are different from the redemption window.

Date is still wrong, because earlier time zones permit 1/1/2024.

PR #4049, opened 2023-11-02
Fix Gen 7 map position

abcboy101:
## Prevent UpdateOverworldCoordinates from corrupting FieldMoveModelSave

Occasionally, changing the player's coordinates was causing the player to ride an invisible Pokémon on load. I determined that this was because FieldMoveModelSave has a uint16 value at offset 0 that appears to be correlated to the player's current Ride Pokémon, not just a copy of the map value. These are the values I observed:

- 1: on foot
- 5: Tauros
- 6: Sharpedo
- 12: Stoutland
- 13: Mudsdale
- 14: Machamp
- 15: Lapras

This doesn't appear to be the only value that determines the player's Ride Pokémon though, since changing it manually either keeps the player on their previous Ride Pokémon or has them mount an immovable, invisible Ride Pokémon. Because of this, I didn't add an accessor or form element for it.

## Fix player coordinates

The player's X/Z/Y coordinates appears to be stored as consecutive float32 values at offsets 0x08/0x0C/0x10, without any bytes in between (rather than 0x08/4-byte gap/0x10/4-byte gap/0x18). X is the left/right axis, Y is the backward/forward axis, and Z is the down/up axis (altitude). I modified the accessors to use the correct offsets.

## Fix player rotation

The player's rotation appears to be stored as a unit quaternion of four float32 values at offsets 0x14/0x18/0x1C/0x20, rather than as a single value. I recorded the following values:

| X | Z        | Y | W        | Direction  | Rotation around z-axis |
| - | -------- | - | -------- | ---------- | ---------------------- |
| 0 | -0.70711 | 0 |  0.70711 | left       |            -π/2 (-90°) |
| 0 | -0.38268 | 0 |  0.92388 | down/left  |            -π/4 (-45°) |
| 0 |  0       | 0 |  1       | down       |             0     (0°) |
| 0 |  0.38268 | 0 |  0.92388 | down/right |             π/4  (45°) |
| 0 |  0.70711 | 0 |  0.70711 | right      |             π/2  (90°) |
| 0 |  0.92388 | 0 |  0.38268 | up/right   |            3π/4 (135°) |
| 0 |  1       | 0 |  0       | up         |             π   (180°) |
| 0 |  0.92388 | 0 | -0.38268 | up/left    |            5π/4 (225°) |

The X and Y components appear to always be 0 (even when the player is on an incline), though it's possible that later games (especially for the camera controls) might actually use them. I don't think SMUSUM does, though.

The player rotated

I made it so that the editor form converts the Z and W components of the quaternion to degrees for general ease-of-use. Having the individual components be editable would also work, but I think a lot of people would find dealing with the quaternion directly confusing.

----

These findings probably have similar implications for how the coordinates in Gen 6/the Switch games are stored, but I haven't looked into them yet.

### DIFF
--- a/PKHeX.Core/Saves/Substructures/Gen7/FieldMoveModelSave7.cs
+++ b/PKHeX.Core/Saves/Substructures/Gen7/FieldMoveModelSave7.cs
@@ -12,5 +12,8 @@ public sealed class FieldMoveModelSave7 : SaveBlock<SAV7>
     public float X { get => ReadSingleLittleEndian(Data.AsSpan(Offset + 0x08)); set => WriteSingleLittleEndian(Data.AsSpan(Offset + 0x08), value); }
     public float Z { get => ReadSingleLittleEndian(Data.AsSpan(Offset + 0x0C)); set => WriteSingleLittleEndian(Data.AsSpan(Offset + 0x0C), value); }
     public float Y { get => ReadSingleLittleEndian(Data.AsSpan(Offset + 0x10)); set => WriteSingleLittleEndian(Data.AsSpan(Offset + 0x10), value); }
-    public float R { get => (int)ReadSingleLittleEndian(Data.AsSpan(Offset + 0x20)); set => WriteSingleLittleEndian(Data.AsSpan(Offset + 0x20), value); }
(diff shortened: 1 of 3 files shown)


kwsch:
Flashbacks to Gen7 release when players thought there was a glitch in the Pokécenter desk due to PKHeX's mishandling of player coordinates https://www.reddit.com/r/pokemon/comments/5eazhi/possible_save_glitch_in_sun_and_moon/

PR #4060, opened 2023-11-12
Naming useful SV blocks

fattard:
I've added names for some common blocks that are useful for my plugin FlagsEditorEX.

I've put the definitions at the same block types.

Feel free to suggest better names.


sora10pls:
To stay consistent with other labeled block keys, I’d advise changing some to KDefeatedTrainers01 and KDefeatedTrainers02 (just keep your comment noting how the latter was added with 2.0.1)

fattard:
I'm still researching about hidden item blocks, looks like there are more blocks reserved for that.
I will add those later.

fattard:
Finished documenting the HIdden Item blocks.
There are a total of 100 blocks with similar structure and initial values.
Currently, only 9 of them are allocated to be used.

There are 4 blocks that get activated during travels in Paldea, they can be identified by covering ~mostly~ of the region mentioned, it's not an exact division.

Then there are other 3 blocks that are activated upon arriving Area Zero, no particular order.

Last 2 identified are activated upon reaching Kitakami.

PR #4065, opened 2023-11-17
Add Alex Dragapult Date

xiaolong11123:


sora10pls:
November 16 is not an eligible date as of the time the serial code was revealed.

xiaolong11123:
Have you calculated the time difference? I'm not sure, so I just subtracted
it by one day.

Matt ***@***.***>于2023年11月17日 周五21:07写道：

sora10pls:
Yes, otherwise I would not have made the comment.

xiaolong11123:
OK, I have changed it.

Lusamine:
For future reference, the math is not to "subtract one from the date." The correct way to know the earliest date is to look for the earliest date anywhere on Earth at the time the serial code is revealed.

PR #4069, opened 2023-11-19
Add GUI For Gen 5 Musical Prop Editing

### LINKED ISSUE #4063: Potentailly Wrong Number Of Props In Musical5
**Describe the bug**

There maybe incorrect information documented for the number of props in the code and thus the final flags assignment may be wrong. On [Bulbapedia Prop Page](https://bulbapedia.bulbagarden.net/wiki/Prop) it lists 100 props and the only place that I saw which has 101 props is [Serebii Musical Page](https://www.serebii.net/blackwhite/pokemonmusical.shtml) but that is a mistake since it lists 2 items twice and is missing the top hat.
https://github.com/kwsch/PKHeX/blob/aec0367085957c6a9e48caa4ea38e3835e9fec45/PKHeX.Core/Saves/Substructures/Gen5/Musical5.cs#L10-L17

This means that the last line of the unlock all props method might needs to be changed to 
```csharp
Data[bitFieldOffset + 0xC] = 0x0F; // top 4 bits unset, to complete multiple of 8 (100=>104 bits).
```

foohyfooh:
Closes #4063

### DIFF
--- a/PKHeX.Core/Resources/text/game/props_bwb2w2/text_props_de.txt
+++ b/PKHeX.Core/Resources/text/game/props_bwb2w2/text_props_de.txt
@@ -1,0 +1,100 @@
+Pink Barrette
+Blue Barrette
+Red Barrette
+Blue Flower
+Crimson Scarf
+Red Flower
+Big Barrette
+Headband
+Square Glasses
+Striped Barrette
+Small Barrette
+Decorative Ribbon
+Purse
+Necklace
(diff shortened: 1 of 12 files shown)


sora10pls:
Thanks for the PR!

Please note that B2W2 text dumps are available out there (as well as tools to dump said text), and proper localizations would be preferred over copy/pasting English for every localization.

PR #4083, opened 2023-12-04
Fix Max Memory Intensity

foohyfooh:
From the saves I have it appears that memory intensity goes from 0 to 7 rather than 0 to 6.


Lusamine:
My Pokemon has an intensity 7 memory.
image
image

Having a memory seems to use intensity 1-7.
Only when there is no memory is intensity 0.

PR #4093, opened 2023-12-07
Update flags_hgss_ja.txt

902PM:
Translated "flags_hgss_ja.txt" into Japanese.


Lusamine:
Thanks for the translations! By the way, if you click on a branch that already exists and edit all the files there, then open a pull request, you can do it with one pull request instead of many. Here are some pictures:

From your repo, select the branch you want to use (fine to use one that's already created).
image

After you've edited all the files on that branch, you can click Contribute > Open Pull Request and it will submit all the changes at once.
image

That way, there will only be one big pull request instead of many like this.
image

PR #4102, opened 2023-12-13
Update WC9 Dates

xiaolong11123:


kwsch:
According to Serebii, the 2 paradox gifts are only going to be valid for one year?
image

PR #4125, opened 2023-12-21
Fix Language Files Not Matching DLC Raid Buttons

foohyfooh:
Fix translation issue caused by PR #4112


kwsch:
Spanish is too long for the button text

foohyfooh:
I had just moved the text from the past location to match the new control name and didn't fix that since the issue would have already existed on master. Is there anything I should do to fix it?

PR #4142, opened 2023-12-30
Add prompt to create subfolders when dumping boxes

FrancoJTrevino:
Adds a prompt that ask if you want to create subfolders for each box when using the option "Dump boxes".
It seems like the logic was already there but it was never used. Also translated the prompt message to spanish.


kwsch:
Closed by https://github.com/kwsch/PKHeX/commit/d25fab0e876815cfd2ee1e8f4bf07666cd5dfb55
I decided it was better to just reimplement behavior with a settings object for flexibility.

PR #4153, opened 2024-01-07
Add Support For SV Throw Style

foohyfooh:


Lusamine:
You were previously advised to use text dumps when available rather than copypasting into every single language: https://github.com/kwsch/PKHeX/pull/4069#issuecomment-1817725397

Text dumps for all of these styles exist. But this looks like it's manually entered with typos, and then you copypasted these errors into every single language.
image

I would suggest getting a text dump and filling in these entries with the proper, official localizations.

foohyfooh:
Sorry about that. It should be fixed now.

foohyfooh:
Send the information for the club room please. I don't think I will look into implementing now for this PR but I may look into implementing it later.

Lusamine:
It would be more appropriate for a separate PR, yes, but here are our notes. There are still some fields that are unknown if you can figure them out.
[clubroom block.txt](https://github.com/kwsch/PKHeX/files/13854034/clubroom.block.txt)

Lusamine:
Btw there is another typo still in "Reverant" as mentioned above in the review.

foohyfooh:
Didn't notice that one sorry.

foohyfooh:
I was working off the data from https://projectpokemon.org/home/forums/topic/64506-request-scarletviolet-change-throw-style-pkhex-block-data/ and I just noticed that the order for the constants and the text don't match up so I need to verify the order.

PR #4173, opened 2024-01-26
Add publicly accessible max file size

codemonkey85:
I need this to set the maximum allowed file size when uploading files to a web app.


kwsch:
Not sure that's a good idea to expose; there's more sizes than just the per-game size whitelists. Don't forget GameCube memory cards!

Have you considered using `FileUtil` to check the size instead?
https://github.com/kwsch/PKHeX/blob/5706fb2fb54282a8f0ac33f3ffbe4e14a3156616/PKHeX.Core/Util/FileUtil.cs#L143-L158

codemonkey85:
@kwsch I can't just check the file size, I have to specify the max file size allowed when opening a read stream to a file in a browser.

`await using var fileStream = browserLoadSaveFile.OpenReadStream(MaxFileSize);`

https://github.com/codemonkey85/PKMDS-Blazor/blob/6d0bcaab1d7f2268bb0f07c982bd2a1e5d96e981/Pkmds.Web/Layout/MainLayout.razor.cs#L60

kwsch:
GameCube memory cards can range from 512KB - 64MB.

I would recommend just using a constant value, say 8MB. Might have issues with large GameCube backups, but it's good enough to match 99.99% of use cases.

kwsch:
I'd say the best practice for this is an arbitrary max as previously mentioned; 64MB vs 8MB shouldn't hurt in the grand scheme of things. Once you have obtained the reference to the `fileStream`, you can check the `fileStream.Length` with the current API of `SaveUtil.IsSizeValid(fileStream.Length)` or just chuck it to `FileUtil.GetSupportedFile` and pivot based on the returned object type.

codemonkey85:
As I said, I have to tell the browser what the max acceptable file size is beforehand, so I can't use the existing utilities to do that. But an arbitrary const works for now.

PR #4175, opened 2024-01-27
add linux compatibility based to version + packages

amidevous:
add linux compatibility based to version + packages


kwsch:
Basic installation instructions for installing the .NET Runtime on various architectures that aren't officially supported are irrelevant to the applications that run on them.

You're more than welcome to create your own repository to store these scripts and alterations; I don't see any need to include them in the application's src repository.

PR #4218, opened 2024-03-13
Add YOASOBI Pawmot Date

xiaolong11123:


sora10pls:
As we've told you before, please verify that your dates are accurate before submitting a pull request.

The event runs from March 13, 2024 to February 28, 2025, with the start date and end date both having tolerance for a day earlier (March 12, 2024) and a day later (March 1, 2025), due to differences in time zones.

PR #4247, opened 2024-04-13
Add Marco's EUIC Iron Hands To wc9.pkl

foohyfooh:
While #4238 added the date information, it didn't add the data to the wc9.pkl file.


Lusamine:
This is nice, but not a requirement for adding dates. WC data is typically bundled at each official release, and users can place the cards in a folder named "mgdb" in the same folder if needed until then.

foohyfooh:
I had just noticed that the past few PRs adding the events had updated the wc9.pkl as well so I had assumed it was an oversight and I had forgotten about the mgdb because of that.

Lusamine:
Historically, Mystery Gifts didn't have date restrictions, so every time a new gift was released, we simply waited until official PKHeX release to add them all. It's new that we're having to document start/end dates.

I think it's preferable that people contributing the dates aren't required to also learn how to add a pkl that's going to be added at the end anyways.

PR #4280, opened 2024-05-23
Fixed SecretBase3Team

Bl4ckSh4rk:
Some data was read from wrong offsets.

### DIFF
--- a/PKHeX.Core/Saves/Substructures/Gen3/SecretBase3.cs
+++ b/PKHeX.Core/Saves/Substructures/Gen3/SecretBase3.cs
@@ -5,6 +5,8 @@ namespace PKHeX.Core;
 
 public sealed class SecretBase3(Memory<byte> raw)
 {
+    public const int SIZE = 160;
+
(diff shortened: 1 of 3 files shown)


PKHeX.Core/Saves/Substructures/Gen3/SecretBase3.cs
```diff
@@ -59,7 +61,7 @@ public ushort SecretBasesReceived
     public Span<byte> GetDecorationCoordinates() => Data.Slice(0x22, 0x10);
     public void SetDecorationCoordinates(Span<byte> value) => value.CopyTo(GetDecorationCoordinates());
 
-    private Span<byte> TeamData => Data.Slice(50, 72);
+    private Span<byte> TeamData => Data.Slice(0x34, 108);
```
kwsch, line 64: What rationale for 108 bytes? The party data struct seems to be 0x48 bytes large:
https://github.com/pret/pokeemerald/blob/18f84b78f2d1a8669753fa586836fca06036c790/include/global.h#L538-L546

PR #4287, opened 2024-06-03
Changes

902PM:
1.Unified with in-game text.(text_cxd_00000_ja.txt)
2.Add Poké Seer's text for clarity.(text_gsc_00000_ja.txt)
3.Fixed some text.(text_rsefrlg_00000_ja.txt)


kwsch:
Seer text should be kept short, because this shows up in drop-downs and displays. It's supposed to be abbreviated rather than an entire long-message.

902PM:
How is it?

PR #4294, opened 2024-06-06
add sophia's gyarados.wc9 to mgdb and rename Kaito Arii's Talonflame actual trainer name

9Bitdo:
- add sophia's gyarados.wc9
- rename Kaito Arii's Talonflame to actual trainer name ナーク


kwsch:
Why is the trainer name of a distributed wc9 changing? The mgdb should not contain manually edited gift binaries.

9Bitdo:
> Why is the trainer name of a distributed wc9 changing? The mgdb should not contain manually edited gift binaries.

the latest Talonflame mystery gift was distributed during the Japan Pokemon Championship and i guess the correct procedure would be to name the wc card as its trainer ? 

as for sophia's gyarados, someone mentioned of the missing wondercard not in the latest pkhex and so i updated it into the wc9.pkl. Please advice if i did incorrectly

sora10pls:
I think there’s a misunderstanding here. The comment associated with the legal met date range for Talonflame was changed, not the actual wonder card data. The only changes made to `wc9.pkl` were adding the Gyarados WC9 data.

Lusamine:
I think the confusing part is that someone else already had a pull request to add the WC https://github.com/kwsch/PKHeX/pull/4292, then you also created a pull request to add the WC and change a comment. Now they have closed their pull request. 

I think it's an issue of etiquette where if someone else already has the pull request, you don't open the same one while adding in something extra and unnecessary. A pull request should do one thing, not sprinkle in extra changes.

9Bitdo:
> I think the confusing part is that someone else already had a pull request to add the WC #4292, then you also created a pull request to add the WC and change a comment. Now they have closed their pull request.
> 
> I think it's an issue of etiquette where if someone else already has the pull request, you don't open the same one while adding in something extra and unnecessary. A pull request should do one thing, not sprinkle in extra changes.

i apologise for the misunderstanding and not trying to mess things up. I will take extra note of this in future

PR #4321, opened 2024-07-14
Update SaveHandlerFooterRTC.cs

IbisLiven:
saves with 0x7 RTC footer generated by [FlashGBX](https://github.com/lesserkuma/FlashGBX) were rejected as too large.
change tested to work on latest release (24.07.03).


kwsch:
Worth mentioning that this is currently an issue with FlashGBX (they're aware); lacks the 64bit `time_t` timestamp which evens it out to the expected 0x10 sized footer.

Might eventually remove this bypass down the road once their update is more generally adopted.

kwsch:
Fixed in FlashGBX latest update.
https://github.com/lesserkuma/FlashGBX/releases/tag/4.1

PR #4340, opened 2024-08-15
Add WCS 2024 Steenee date

9Bitdo:


Lusamine:
Both your start and end dates are incorrect.

The codes were given out early and are currently redeemable which means they are at minimum legal today (August 15, 2024). I am not sure if they were redeemable at a time that would have qualified for August 14, 2024.

The cards say end of redemption is August 30, 2024 at 23:59 UTC. Here is a picture of a card with the end date:
image

Lusamine:
Also, I think there may be an issue with your pkl that you included because it's smaller. Maybe you deleted older events somehow.

9Bitdo:
> Both your start and end dates are incorrect.
> 
> The codes were given out early and are currently redeemable which means they are at minimum legal today (August 15, 2024). I am not sure if they were redeemable at a time that would have qualified for August 14, 2024.
> 
> The cards say end of redemption is August 30, 2024 at 23:59 UTC. Here is a picture of a card with the end date: image



my bad on the inaccurate end date,and the .pkl i will closed the PR . At this point of time i believe the redemption date would be the best accurate as what the screenshot of the redemption card state, since honolulu is gmt -10 , i believe the min legal date will be as 15th Aug since -11 region would still fall on the aug 15th. Thank you and i relook into the errors before resubmitting

PR #4359, opened 2024-09-23
Support G2 mail in all languages

### LINKED ISSUE #3470: Kenya Mail not being read/written correctly in Gold/Silver (Unknown if Crystal as well)
Kenya's mail will break and no longer be recognised by the game as valid if either Kenya or the Mailbox is interact with in any way. The letter cannot be manually fixed in PKHeX, and the only workaround is to either reset the "obtained" flag and get another Kenya/letter,  or restore to an earlier unedited save. 

**To Reproduce**
1. Interact with Kenya in a way that would cause its mail to be read in PKHeX (either by switching Kenya's position in the party, or by viewing and then saving the mail submenu (viewing the letter specifically does not make a difference either way, but if the letter is viewed, the "t" in "to another path" is not displayed. 
2. Attempt to interact with the man on Route 31, who will claim it is not his letter. Editing the "t" back into the letter has no effect on whether or not he'll accept it, and multiple bytes are changed. 
3. If Kenya is moved to another slot, the slot for the mailbox associated with the party slot will not update accordingly. However, I've been unable to check if this has any effect on whether or not the letter will be accepted, and the mail being in the correct slot will also result in an invalid letter anyway. Manually editing a letter into the correct position with the correct details also does not result in a legal letter. 
4. Regardless of the method used to reproduce this glitch or whether or not you attempt to restore the message, the second line of the letter will be blanked out entirely, and the NPC will not accept it. 

**Expected behavior**
PKHeX would presumably not break any letters in general. I've been unable to replicate this bug with other letters created by the player, so something about this one causes it to be incorrectly read. In addition, multiple bytes are changed upon the program "viewing" this letter even tangentially. These bytes again wildly differ if one attempts to "correct" the letter. 

**Screenshots**
mail
addresses

**Additional context**
Enclosed save files show unaltered and altered bytes for the letter with the bug replicated by viewing the mail tab and saving it. Relevant addresses are 0x6FA-6FB, 0x70B, and 0x712-715 (the same information also seems to be present at 0x814-815, 0x825, 0x82C-82F).
[saves.zip](https://github.com/kwsch/PKHeX/files/8357111/saves.zip)

Hope this helps, will be happy to submit further information upon request.

### LINKED ISSUE #4027: GSC Box names not decoded correctly
**Describe the bug**
In Gold, Silver, and Crystal, there is a wider selection of characters that can be entered in Box names than in Trainer names and Pokémon nicknames. In English, the additional characters include é and lowercase letters preceded by an apostrophe.

PKHeX does not correctly render Box names for Gold, Silver, and Crystal if they use characters that cannot be entered in nicknames, because it uses the Generation 1 encoding for both Generation 1 and 2.

**To Reproduce**
Example 1:
1. When entering a Box name in Gold, Silver, and Crystal, switch to lowercase to access the additional characters.
2. Name a box "Pokémon"
3. In PKHeX, this Box's name displays as "Pok"

Example 2:
1. When entering a Box name in Gold, Silver, and Crystal, switch to lowercase to access the additional characters.
2. Name a box "I'm you're" (using the special combined apostrophe + letter characters)
3. In PKHeX, this Box's name displays as "Iñm youòe"

**Expected behavior**
For é, this character should display as é, and not cut off the rest of the box name. (Personally, I would suggest that invalid characters should display as � and should not prevent the rest of the string being decoded, so that even with the current encoding map it would display as "Pok�mon", but I can understand there may be good reasons to stop decoding at the first invalid character.)

For the apostrophe + letter characters, while they definitely shouldn't be rendering as the wrong characters, I have no suggestions for how to represent them as a single character. Especially since a standalone apostrophe followed by a lowercase letter is just as valid in a Box name as the combined character. (FWIW Bank just renders them as the letter without the apostrophe for the English ones, and for the French ones it doesn't even try and just prints just a space.)

**Additional context**
The character encoding for characters that cannot appear in nicknames or Trainer names differs between Generation 1 and 2 (and between languages). PKHeX is currently assuming it can use the Generation 1 encoding for all Western Generation 1 and 2 games, but this assumption does not hold for Box names or Mail because of the characters available.

In other Western languages, you can type ß and various diacritic characters in Box names (the available characters [depend on the language](https://bulbapedia.bulbagarden.net/wiki/Text_entry_(Generation_II))), with the German & French and Spanish & Italian character encodings being incompatible with both each other and their corresponding Generation 1 variants for many of these characters. Even é is located at different code points between Gen 1 and 2 (as evidenced by the issue above).

This almost certainly affects Mail too, since Mail has an even bigger keyboard than Box names, but I didn't specifically check it. Mail also has a different character set depending on the language of the game that wrote the Mail, with overlapping codepoints (e.g. French can type 'ç' in Mail, which is stored at the same code point as Á in Italian & Spanish, but all Western games except English GS can successfully read 'ç' if the Mail was typed in a French game).

abcboy101:
### Handle Japanese/Korean mail
Japanese and Korean mail does not include the "nationality" field, and only store the author's name as 5 characters. The Korean encoding is double-byte, so both the message and author use twice as many bytes.

----

### Handle both NPC and user-entered mail
User-entered mail in Japanese/EFIGS always puts the line break in the same place, and pads it with 0x50 up to the end:
```
XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX 4E | line 1
XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX    | line 2
```

User-entered mail in Korean can have a line break anywhere in the first line. It still pads it with 0x50 up to the end if needed:
```
[No text entered]
59 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 | line 1
50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50 50    | line 2

[Full text entered]
XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX 59 | line 1
XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX    | line 2
```

On the other hand, the message in Randy's mail is too short in every language. In Japanese/EFIGS, it [copies 0x20 bytes](https://github.com/pret/pokecrystal/blob/7b5986006f6b325e471fee25903c769ce67f5da9/engine/overworld/scripting.asm#L1704) from the ROM into a [buffer of size 0x21](https://github.com/pret/pokecrystal/blob/7b5986006f6b325e471fee25903c769ce67f5da9/ram/wram.asm#L2016), meaning that whatever value in WRAM was previously there is not overwritten and is [copied to the mail](https://github.com/pret/pokecrystal/blob/7b5986006f6b325e471fee25903c769ce67f5da9/engine/pokemon/mail.asm#L210).
```
[Japanese]
B8 D7 D4 D0 25 CE D7 B1 C5 7F B6 D7 CA 4E | "くらやみ※ほらあな　からは"
3D C2 25 D0 C1 1D C2 32 B2 C3 B2 D9 D6 50 | "ベつ※みち※つづいているよ"
CB B7 E0 B8                               | Next in ROM: "ひきゃく"

[English]
83 80 91 8A 7F 82 80 95 84 7F AB A4 A0 A3 B2 4E | "DARK CAVE leads"
B3 AE 7F A0 AD AE B3 A7 A4 B1 7F B1 AE A0 A3 50 | "to another road"

[French]
80 8D 93 91 84 7F 8D 8E 88 91 84 7F 7F 7F 7F 4E | "ANTRE NOIRE    "
AC BB AD A4 7F A0 B4 B3 B1 A4 7F AF A0 B1 B3 50 | "mène autre part"

[German G/S]
85 88 8D 92 93 84 91 87 C1 87 8B 84 7F 7F 7F 4E | "FINSTERHÖHLE   "
A8 B2 B3 7F A4 A8 AD 7F 8F A5 A0 A3 7F 7F 7F 50 | "ist ein Pfad   "

[German Crystal]
83 94 8D 8A 84 8B 87 C1 87 8B 84 7F 7F 7F 7F 4E | "DUNKELHÖHLE    "
A8 B2 B3 7F A4 A8 AD 7F 8F A5 A0 A3 7F 7F 7F 50 | "ist ein Pfad   "

[Italian]
86 91 E8 92 82 94 91 80 7F B5 A0 7F A0 A3 7F 4E | "GR․SCURA va ad "
B4 AD 7F A0 AB B3 B1 AE 7F 8F 84 91 82 E8 7F 50 | "un altro PERC․ "

[Spanish]
82 94 84 95 80 7F 8E 92 82 94 91 80 7F A3 A0 4E | "CUEVA OSCURA da"
A0 7F AE B3 B1 AE 7F A2 A0 AC A8 AD AE 7F 7F 50 | "a otro camino  "
```

In Korean, it fills the entire message buffer with data from the ROM.
```
[Korean]
06EE 0332 0797   7F 031F 018C 0701 05DD 052E 099D   59    | "어둠의 동굴에서부터"
02D9 0435 01B6 079C   7F 070C 0141 0327 06EE 07A6 06EE 50 | "다른길이 연결되어있어"
0669 052E 0437 020B 50                                    | Next in ROM: "심부름꾼"
077F 0826 50                                              | Next in ROM: "유지"
6B 48 31 2A 00 09 B3 5B 4D                                | Next in ROM
```

To determine if it's user-entered, we can check that Japanese/international mail has the line break at 0x10, or that Korean mail has no trash bytes past the first terminator. If it isn't user-entered, we preserve any trash bytes on write for the message/author name. This fully fixes #3470.

----

### Display European mail based on mail language

The nationality field of European mail stores the language it was written in, since Crystal and non-English G/S support displaying mail in other languages. Instead of using the game's language, we use the mail's language to display it instead. This fully fixes #4027.

When European mail is sent from Crystal or non-English G/S, it is [converted](https://github.com/pret/pokecrystal/blob/7b5986006f6b325e471fee25903c769ce67f5da9/engine/pokemon/european_mail.asm) to a language-unaware encoding that displays more characters correctly in English Gold/Silver. When it is received back in Crystal or non-English G/S, it is converted back to the original language-specific encoding. On load/save, we perform this remapping in reverse on foreign mail in English G/S so that the editor displays it as it would have in the origin language.

### DIFF
--- a/PKHeX.Core/PKM/SK2.cs
+++ b/PKHeX.Core/PKM/SK2.cs
@@ -78,7 +78,7 @@ public sealed class SK2 : GBPKM, ICaughtData2
         }
     }
 
-    // 0x1F
+    public byte HeldMailID { get => Data[0x1F]; set => Data[0x1F] = value; }
 
     public byte PokerusState { get => Data[0x20]; set => Data[0x20] = value; }
     // Crystal only Caught Data
--- a/PKHeX.Core/PKM/Strings/StringConverter1.cs
+++ b/PKHeX.Core/PKM/Strings/StringConverter1.cs
@@ -209,7 +209,7 @@ public static class StringConverter1
     internal const char MNY = '¥'; // Yen
     internal const char LPO = '@'; // Po
     internal const char LKE = '#'; // Ke
-    internal const char LEA = '%'; // é for Box
+    internal const char LEA = '%'; // é for Box/Mail
     public const char DOT = '․'; // . for MR.MIME (U+2024, not U+002E)
     internal const char SPF = '　'; // Full-width space (U+3000)
     public const char SPH = ' '; // Half-width space
@@ -229,7 +229,7 @@ public static class StringConverter1
         'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', // A0-AF
         'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', 'à', 'è', 'é', 'ù', 'À', 'Á', // B0-BF
         'Ä', 'Ö', 'Ü', 'ä', 'ö', 'ü', 'È', 'É', 'Ì', 'Í', 'Ñ', 'Ò', 'Ó', 'Ù', 'Ú', 'á', // C0-CF
-        'ì', 'í', 'ñ', 'ò', 'ó', 'ú', NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, '←', '\'', // D0-DF
+        'ì', 'í', 'ñ', 'ò', 'ó', 'ú', 'º', NUL, NUL, NUL, NUL, NUL, NUL, NUL, '←', '\'', // D0-DF
         '’', LPK, LMN, '-', NUL, NUL, '?', '!', DOT, '&', LEA, '→', '▷', '▶', '▼', '♂', // E0-EF
         MNY, '×', '.', '/', ',', '♀', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', // F0-FF
     ];
@@ -242,8 +242,8 @@ public static class StringConverter1
         'だ', 'ぢ', 'づ', 'で', 'ど', NUL, NUL, NUL, NUL,  NUL, 'ば', 'び', 'ぶ', 'ベ', 'ぼ', NUL, // 30-3F
         'パ', 'ピ', 'プ', 'ポ', 'ぱ', 'ぴ', 'ぷ', 'ペ', 'ぽ', NUL, NUL, NUL, NUL, NUL, NUL, NUL, // 40-4F
         NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, TOT, NUL, NUL, // 50-5F
-        NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, // 60-6F
-        '「', '」', '『', '』', '・', '⋯', NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, SPF, // 70-7F
+        NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, NUL, 'ぃ', 'ぅ', // 60-6F
+        '「', '」', '『', '』', '・', '⋯', 'ぁ', 'ぇ', 'ぉ', NUL, NUL, NUL, NUL, NUL, NUL, SPF, // 70-7F
         'ア', 'イ', 'ウ', 'エ', 'オ', 'カ', 'キ', 'ク', 'ケ', 'コ', 'サ', 'シ', 'ス', 'セ', 'ソ', 'タ', // 80-8F
         'チ', 'ツ', 'テ', 'ト', 'ナ', 'ニ', 'ヌ', 'ネ', 'ノ', 'ハ', 'ヒ', 'フ', 'ホ', 'マ', 'ミ', 'ム', // 90-9F
         'メ', 'モ', 'ヤ', 'ユ', 'ヨ', 'ラ', 'ル', 'レ', 'ロ', 'ワ', 'ヲ', 'ン', 'ッ', 'ャ', 'ュ', 'ョ', // A0-AF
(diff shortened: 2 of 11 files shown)


kwsch:
Don't worry about the merge conflicts; I can resolve them.

As time goes on, I'm starting to think that it would be better handling these language-isolated games as deriving from a base abstraction rather than a shared class; all the way up to SAV1 and SAV2 having their own derived classes for each of the language variants (KOR, JPN, INT). But that's a lot of work for not much gain besides making future improvements for Gen1/2 easier, when time can be spent on more interesting things :)

abcboy101:
Just added support for Stadium 2 mail as well. Stadium 2 can store 50 mail in the Mailbox, plus up to 30 mail held by Pokémon in the N64 Boxes, using a held mail ID at offset 0x1F in the SK2 format.

For now, the Batch Editor can be used to view/edit the held mail IDs, but I haven't implemented any validation for those values.

PR #4362, opened 2024-09-25
Labels for Gen 5 Trainer Records (plus renamed flag)

WonderSquid:
The records stuff is probably really, really bad, and I hope what I have here is at least semi-useable. Apologies in advance. 

Also relabelled an incorrectly labelled flag on every language but Chinese, since there seemed to be some confusion as to what the trade trainer in Nacrene is named on the wiki, and I don't have a Chinese copy of the game to go check if the wiki is wrong or if the flag name is.


WonderSquid:
please do not look directly at my spaghetti code i'll figure something out

Lusamine:
> Also relabelled an incorrectly labelled flag on every language but Chinese, since there seemed to be some confusion as to what the trade trainer in Nacrene is named on the wiki, and I don't have a Chinese copy of the game to go check if the wiki is wrong or if the flag name is.

If you don't know what the flag should be named, I would suggest reverting it to English and let someone who speaks Chinese translate it. Focus on making the PR work, and don't let the translation issue hold it up.

PR #4436, opened 2025-02-03
Handle Arbok Mainland China wondercard

Manu098vm:
Unlike the previous China Mainland-exclusive wondercard (Meltan), Arbok has preset OT values for all languages. 

immagine

Although it is not legally possible to redeem this wondercard in any language other than CHS, PKHeX currently sets the `redeemLanguage` based on the active save file. This causes OT and Nickname info to be generated according to the save file’s language instead of CHS.

For Arbok, this results in it being generated with the Nickname flag on any non-CHS save. This can also cause problems for ALM, which fails to pull the correct encounter unless "Language: 9" is manually specified in the set.

Since this wondercard can only be redeemed on CHS save files, the `redeemLanguage` should always be CHS, not just the `language` of the generated entity.

### DIFF
--- a/PKHeX.Core/MysteryGifts/WB7.cs
+++ b/PKHeX.Core/MysteryGifts/WB7.cs
@@ -319,9 +319,6 @@ public sealed class WB7(byte[] Data)
 
     public int GetLanguage(int redeemLanguage)
     {
-        if (IsMainlandChinaGift)
-            return (int)LanguageID.ChineseS;
-
         var languageOffset = GetLanguageIndex(redeemLanguage);
         var value = Data[0x1D8 + languageOffset];
         if (value != 0) // Fixed receiving language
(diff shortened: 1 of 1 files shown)


PKHeX.Core/MysteryGifts/WB7.cs
```diff
@@ -315,13 +315,10 @@ public override string OriginalTrainerName
     /// <remarks>
     /// Gifts received can be locally traded to the international release, so there is no need to consider residence or transferability.
     /// </remarks>
-    public bool IsMainlandChinaGift => CardID is (1501);
+    public bool IsMainlandChinaGift => CardID is (1501 or 1503);
```
kwsch, line 318: Note for future-us: if they add any more events, we can likely relax this range check to

```cs
CardID is (> 1500 and <= 1550);
```
or something, just so that it doesn't need to be revised for a 4th+ card :D

PR #4440, opened 2025-02-06
Add Pokémon Day 2025 Flying Tera Type Eevee's date

9Bitdo:
Official redemption would be made possible on 7 Feb onward (if we uses NZDT +13 would mean 6 Feb for =＜ UTC+3 Time zone). Considering only a niche of users that would have access to a redemption code before 7 Feb( Kurt mentioned on 5 Feb (UTC+8) that redemption could be made at least possible on the 4 Feb but probably better to have a reasonable start date), making 5 Feb as a start date for those with access to the code early would be fair and within an acceptable tolerance. Redemption code for the distribution must be made by 30 June 2359 (PDT hence UTC+12 time zone would still be able to redeem it on the 1 Jul 2025.

### DIFF
--- a/PKHeX.Core/Legality/Encounters/Data/Live/DistributionWindow.cs
+++ b/PKHeX.Core/Legality/Encounters/Data/Live/DistributionWindow.cs
@@ -1,0 +1,42 @@
+using System;
+
+namespace PKHeX.Core;
+
+/// <summary>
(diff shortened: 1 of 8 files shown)


9Bitdo:
corrected: based on time zone New Zealand(NZDT +13) would be the earliest to redeem

PR #4444, opened 2025-02-08
Check Egg Move combinations for a valid father

### LINKED ISSUE #4426: Gen 2-5 Allow Combinations of Egg Moves
**Describe the problem**
I noticed that if you have a Pokemon with Egg Moves that are only possible from different sets of fathers (Gen 3 Blastoise with Haze and Flail), PKHeX sees this as legal solely due to them being Egg Moves. From what I understand, starting in Gen 6, both parents can pass down Egg Moves, allowing combinations like this to occur. 

**Expected behavior**
Egg Moves from different sets of fathers should be impossible in Gen 2-5 and seen as illegal.

abcboy101:
Closes #4426

- Both legality checks and generated encounters with requested moves are verified through this class. Repeated lookups are cached.
- When called, the method constructs a graph, and uses a breadth-first search to look for a chain of fathers that can collectively learn the needed moves by level up, and pass them down as Egg Moves all the way to the Egg.
  - If that doesn't work, it then checks encounter data for each father for any transfers, static encounters, etc. that can learn all of the needed moves.
  - If no matching encounter is found, the combination is deemed impossible.
- Currently, the chain is only logged to the console during debug, but it might be nice to display it in the detailed legality info.

----

One of the existing test cases appears to have an invalid combination based on this check. This Gen2 [Oddish](https://github.com/kwsch/PKHeX/blob/9dedd336946dc72b94ab6fa05d0a96a222cbe7d8/Tests/PKHeX.Core.Tests/Legality/Legal/Generation%202/043%20-%20ODDISH%20-%202DCE.pk2) has Flail/Razor Leaf/Swords Dance/Synthesis.
- The only father who can pass Flail to Oddish is the Chikorita line, so the father must be a member of the Chikorita line with Flail/Razor Leaf/Swords Dance/Synthesis.
- For Chikorita to have Flail, it must be chain bred via the Squirtle line as the direct father. But, Chikorita can only get Swords Dance as an Egg Move, and the Squirtle line can't learn it. Therefore, the combination is impossible.


kwsch:
Potential observations/insights:

1. Form appears to be irrelevant, as form-specific movesets/eggmoves didn't become a "thing" until Gen7 with Alolan forms. Can remove it from all bits of the code? The only exception could be Wormadam, but Burmy has no egg moves.
2. Female-only breeds should consider inherited father moves (level-up, TM) as if they were egg moves.
3. Count==1 should be a special case similar to count==0. Is always valid unless it's a female-only breed, as we've already removed any impossible egg moves. This can skip a lot of computation for even more cases, but I guess we can have a setting bool to "be detailed" and check anyway (to find which father would source).
4. API signatures: should be possible to define a method to search for a chain with only `species, moves[], version`. Useful for unit testing. `IsPossible` already satisfies this (sans `Form`), but is private.
5. Smeargle: Can know any move. The API doesn't need to check if it's a sketch-able move, because in Gen2-5, all egg/levelup moves to be checked (Struggle will always be flagged, and Chatter is always legal on Chatot via two Chatot parents and thus won't ever be needed to check).
6. More often than not, egg move inheritance will not require a complex/deep verification. There is a LOT of setting up being done for every time the verifier is called. Is there opportunity for general precomputation (pkl)? The encounter database is "complete", so a leaf-node list of "special" moves & species that learn them could be useful. `(move, species[])[]`. When a chain search fails to find the next depth, we can search for all remaining moves to see if any single species can be specially obtained with all.
7. Result data structure: With above mentioned distillations, it appears that a chain node can be expressed as a 16-bit value comprised of 4 move bitflags and the lower bits for species. For a 4-move inheritance, we can store this in a single `ulong` struct, and thus be allocation free without needing any allocation or stack data structure.
8. Could recursion (to mutate the ulong) be a better approach? Bitflags to indicate the un-appended indexes of the span.
9. There are differences in learnsets/eggmoves between versions of the same context; I don't know if there's any "unlocked via crossover" possibilities that aren't considered in this version-specific approach.

Historically, the egg-chain theorycrafting was done by dumping a list of Move->species[]. This approach might be useful, if we also want to precompute seedpoke[species] (baby eggs from current species) and (group1, group2)[species]. We're already doing the work when the node list is built... so if we only have to do some of the work once and keep a global copy...

Maybe a rent on `HaveVisited bool[species]` and `MovesPassable ushort[species` can give us some lazy computation with recursion, as the recursion will update a ulong struct until bits are all cleared or the search ends. No need to allocate aside from the initial setup for each context (2,3,4,5).

abcboy101:
1-3 and 6 should be addressed for now, the rest I'll try to get to another day.

For 2, I'll note that only inherited level-up moves are currently considered, since it's only directly checking movesets in the current generation anyway (past-gen TMs are handled through attempted encounter generation for now).

For 3, there were some impossible Egg Moves in GS that needed to be removed:
- Bulbasaur, Oddish, Snorlax (Charm)
- Staryu (Aurora Beam, Barrier, Supersonic)
- Smoochum (Lovely Kiss)

For 5, [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Sketch_(move)) and [Showdown](https://github.com/smogon/pokemon-showdown/blob/master/data/mods/gen2/moves.ts) claim that depending on the game, certain moves (Mirror Move, Sleep Talk, Transform, Metronome, Mimic, Self-Destruct and Explosion) also can't be Sketched. They don't agree on whether Mirror Move or Sleep Talk can't be Sketched or not in Gen 3-5 though (Bulbapedia claims they can't, Showdown claims they can).

Lusamine:
> They don't agree on whether Mirror Move or Sleep Talk can't be Sketched or not in Gen 3-5 though (Bulbapedia claims they can't, Showdown claims they can).

Bulbapedia currently says these two moves can be Sketched since Gen 3 which would agree with Showdown.

image

It's not immediately clear to me if any of these moves will matter for father egg move legality but they should be considered for Smeargle move legality, which is a separate issue.

There were still a couple of legality issues for this PR that were not handled and I will have to find them again.

Tests/PKHeX.Core.Tests/Legality/LearnabilityTests.cs
```diff
@@ -58,4 +59,49 @@ public static void VerifyCanLearnTM(string species, params string[] moves)
         var can = EncounterLearn.CanLearn(species, moves);
         can.Should().BeTrue($"{species} should be able to learn all moves: {string.Join(", ", moves)}");
     }
+
+    [Theory]
```
andrebastosdias:
```suggestion
    [Theory]
    [InlineData(GameVersion.GS, Species.Paras, Move.Counter, Move.Flail, Move.LightScreen)]
```

https://www.smogon.com/forums/threads/bug-report-validator.3774996/
> Magikarp learns Flail by levelup.
> Chainbreeding Flail: Magikarp -> Horsea ->Squirtle -> Chikorita -> Paras -> Heracross.
> Heracross learns Counter by levelup.
> Chainbreeding Flail+Counter: Heracross -> Chikorita.
> Chikorita learns Light Screen by levelup.
> Breeding Chikorita with Light Screen, Counter and Flail with female Paras results in Paras having those moves.

PR #4446, opened 2025-02-14
Handle HOME Keldeo & Meltan; Revise WB7

Manu098vm:
The AbilityType for LGPE wondercards was hardcoded to 3, to ensure that the ability of generated entities was picked randomly (Any12). From what I’ve seen, the Meltan distributed via the HOME gift always has Ability 1. To make sure the first ability slot is always chosen during generation, the program needs to properly read the AbilityType from the wondercard. I’ve set the wondercard to contain AbilityType 0, so that 'OnlyFirst' is selected.

The handling for height, weight, and scale follows the same approach as the other classes.

Meltan is particularly tricky. The mystery gift stores the Souvenir Ribbon in the core data of the PH3. However, the WB7 and PB7 classes do not currently handle ribbons. It might be possible to utilize unused space within the WB7 and PB7 data, but the code flow would need to be reworked. I’m not sure how we should proceed to properly account for that ribbon.

### DIFF
--- a/PKHeX.Core/Legality/Verifiers/MiscVerifier.cs
+++ b/PKHeX.Core/Legality/Verifiers/MiscVerifier.cs
@@ -642,15 +642,16 @@ public sealed class MiscVerifier : Verifier
             data.AddLine(GetInvalid(LDateOutsideConsoleWindow, Misc));
     }
 
-    private static void VerifyAbsoluteSizes(LegalityAnalysis data, IScaledSizeValue obj)
+    private static void VerifyAbsoluteSizes<T>(LegalityAnalysis data, T obj) where T : IScaledSizeValue
     {
-        if (data is { Entity: PB7 , EncounterMatch: WB7 { IsHeightWeightFixed: true } enc })
-            VerifyFixedSizes(data, obj, enc);
+        if (obj is PB7 pb7 && data.EncounterMatch is WB7 { IsHeightWeightFixed: true } enc)
+            VerifyFixedSizes(data, pb7, enc);
         else
             VerifyCalculatedSizes(data, obj);
     }
 
-    private static void VerifyFixedSizes(LegalityAnalysis data, IScaledSizeValue obj, WB7 enc)
+    // ReSharper disable 4 CompareOfFloatsByEqualityOperator -- THESE MUST MATCH EXACTLY
+    private static void VerifyFixedSizes<T>(LegalityAnalysis data, T obj, WB7 enc) where T : IScaledSizeValue
     {
         if (obj.HeightAbsolute != enc.GetHomeHeightAbsolute())
             data.AddLine(GetInvalid(LStatIncorrectHeight, Encounter));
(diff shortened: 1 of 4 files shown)


Manu098vm:
About the Ribbon Issue:
Afaik, ribbons are actually stored in PB7 data when a Pokémon is transferred from HOME to LGPE. See how the Souvenir Ribbon is retained when transferring the Meltan HOME -> LGPE -> HOME -> SwSh. 

Hence, the simplest solution may be to implement proper ribbon handling—along with transfer logic to other games—at least for the Souvenir Ribbon, in both the WB7 and PB7 classes.

Other ribbons wouldn't be strictly necessary, as there’s no way to obtain them in a PB7 file.

This approach would ensure that we have a legal PB7 containing the correct ribbon data, and also allows us to check ribbon legality in Gen8+ games, even if the original data comes from a WB7.

I can eventually look into it in the coming days, but it'll probably be later than your planned eta timeline.

Manu098vm:
Another thing to account for with this Meltan:

Not only it has a set HeightScalar and WeightScalar (128), it also has fixed HeightAbsolute and WeightAbsolute values in the DataPB7 of the PH3. See the screenshot below:

<details> <summary>See Screenshot</summary> Meltan DataPB7 </details>

(The values are consistent among other redemption/dumps)

These values don’t match the correctly calculated ones from the Personal Info, which causes PKHeX to raise a legality flag. These values are retained upon transfer into LGPE.

As suggested by Anubis, after transferring it to LGPE, I tested the following scenarios to see if the game would fix/recalculate HeightAbsolute/WeightAbsolute (copy-paste from the discord):

<details> <summary>Test Results</summary>

- Healing at a Pokémon Center – No change
- Placing in party – No change
- Marking as favorite in party – No change
- Participating in battle (not as active Pokémon) – No change
- Participating in battle (as active Pokémon) – No change
- Catching another Pokémon – No change
- Registering a new Pokémon in the Pokédex – No change
- Feeding Meltan a Meltan Candy – No change
- Leveling up with a Rare Candy – No change
- Removing from party – No change
- Depositing in daycare – No change (checked from the daycare section in PKHeX)
- Retrieving from daycare – No change
- Doing Pro Training – No change
- Taking out of Poké Ball – No change
- Putting back in Poké Ball – No change
- Sending it back to HOME after leveling up and powering up in LGPE – No change
- Transferring to another save file via HOME – No change
- Trade from LGPE games – No change
- Trade from Home accounts, one trade an untouched one, the other trade the leveled up Meltan; both – No change
</details>

In the end, nothing changed the values, so I can say that no regular gameplay mechanics trigger a recalculation of HeightAbsolute and WeightAbsolute. PKHeX will need to account for these incorrect values. The following commit will address this issue.

With this, the only remaining issue (strictly regarding this Meltan), is handling the Souvenir Ribbon.

Note: The Souvenir Ribbon remains present even after transferring Meltan from HOME -> LGPE -> HOME. It can be viewed in SwSh, and also in HOME itself if the Pokémon is sent back to HOME after visiting SwSh. Anubis found the offset that stores the Souvenir Ribbon flag in the PB7.

PR #4476, opened 2025-04-21
Updated

DaveTmire85:
Made minor code optimizations, updated support for the latest Windows targets

### DIFF
--- a/PKHeX.WinForms/PKHeX.WinForms.csproj
+++ b/PKHeX.WinForms/PKHeX.WinForms.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
-    <TargetFramework>net9.0-windows</TargetFramework>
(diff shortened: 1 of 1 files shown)


DaveTmire85:
I've built the Windows version, and I'm expanding on the UI quite a bit. 

**FOR THE FUTURE:**

My modernization plan for PKHeX focuses on implementing a contemporary design system like MaterialSkin.2 or Krypton Toolkit to replace the dated Windows Forms interface. The proposal recommends a phased approach starting with a theme engine that supports dark mode, followed by gradual replacement of standard controls with modern alternatives featuring animations, depth effects, and improved visual hierarchy. Key improvements will include a card-based layout for Pokémon data, interactive stat visualizations, a sidebar navigation replacing the current tab system, and toast notifications instead of message boxes. This transformation would preserve all PKHeX's functionality while delivering a more intuitive, visually appealing experience aligned with modern application design standards.

PR #4478, opened 2025-04-21
Update MemeKey Against Encryption with ECB Vulnerability

Xieon:
ECB should not be used as a mode for encryption. It has dangerous weaknesses. Data is encrypted the same way every time meaning the same plaintext input will always produce the same ciphertext. This makes encrypted messages vulnerable to replay attacks.

Recommendation
Use a different CypherMode.

References
Wikipedia, Block cypher modes of operation, Electronic codebook (ECB). Common Weakness Enumeration: CWE-327.

First detected in commit on Sep 16, 2024
PKHeX.Core/Saves/Encryption/MemeCrypto/ MemeKey.cs:134 on branch master

Potential Fix: 
Potential fix for Encryption using ECB

To fix the issue, we need to replace the use of CipherMode.ECB with a more secure mode, such as CipherMode.CBC (Cipher Block Chaining). CBC mode requires an initialization vector (IV) to ensure that identical plaintext blocks produce different ciphertext blocks, enhancing security. The IV should be randomly generated for each encryption operation and transmitted alongside the ciphertext (e.g., by prepending it to the ciphertext).

The changes involve:

Modifying the GetAesImpl method to use CipherMode.CBC instead of CipherMode.ECB. Generating a random IV for encryption and ensuring it is used during decryption. Updating the encryption and decryption logic to handle the IV appropriately.

Actual Fix: 
The changes update the AES encryption and decryption logic in the MemeKey class to use Cipher Block Chaining (CBC) mode instead of Electronic Codebook (ECB) mode. The AesEncrypt and AesDecrypt methods now utilize CBC mode with an explicitly initialized zeroed IV. The method responsible for creating the AES implementation is also modified to set the cipher mode to CBC and initialize the IV accordingly. No changes were made to method signatures or the overall class interface; only the internal cryptographic implementation was altered.

Changes
File(s)	Change Summary
PKHeX.Core/Saves/Encryption/MemeCrypto/MemeKey.cs	Updated AES encryption/decryption from ECB to CBC mode, initializing IV to zeroes within methods.


Xieon:
Ill play around with it later - but I'm sure you know about it, and its likely not any type of big deal. I guess I don't have build in my actions just code check - it should've failed for my build and I wouldn't have pushed it.  Sorry -

Lusamine:
It looks [like this is AI generated](https://github.com/Xieons-Gaming-Corner/XGC_PKHeX/pull/60), and the AI has no idea what this code is for. Your "fix" would break all gen 7 save files. Please do not submit any more pull requests of this nature.

PR #4479, opened 2025-04-23
Add Pokémon Town - Ditto Project (KOR) date

bdawg1989:
Add date for (https://github.com/projectpokemon/EventsGallery/commit/8241a1fcb589cb1b0bb85002b0eee1965ef7bb27)

### DIFF
--- a/PKHeX.Core/Legality/Encounters/Data/Live/EncounterServerDate.cs
+++ b/PKHeX.Core/Legality/Encounters/Data/Live/EncounterServerDate.cs
@@ -217,6 +217,7 @@ public static class EncounterServerDate
         {0514, new(2025, 02, 05, 2025, 07, 01, +2)}, // Pokémon Day 2025 Flying Tera Type Eevee
         {0519, new(2025, 02, 20, 2025, 03, 01)}, // Marco's Jumpluff
         {0066, new(2025, 04, 18, 2025, 08, 01)}, // Wei Chyr's Rillaboom
+        {1019, new(2025, 04, 24, 2025, 05, 07)}, // Pokémon Town - KOR Ditto Project
 
         {9021, HOME3_ML}, // Hidden Ability Sprigatito
         {9022, HOME3_ML}, // Hidden Ability Fuecoco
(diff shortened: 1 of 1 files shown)


Lusamine:
Valid redemption date looks incorrect: https://x.com/metamongproject/status/1914619520377090071?t=ICqZWHPkdIEl3MWVZrgRSQ
Please refer to official notices and make sure you are accounting for timezones.
(Yes, this means if you have to go to a mall to get a code, you need to verify the mall's time of opening.)

bdawg1989:
Based on the Image provided on X, I have updated the dates.

Event dates and locations:

April 25 (Thursday) ~ May 18 (Saturday) at Lotte World Adventure Outdoor Plaza
May 1 (Thursday) ~ May 6 (Tuesday) at Gangnam Lotte World Tower 2nd Floor (near ZARA)

Lusamine:
Your dates are still incorrect. I suggest you explain your reasoning for why you decided on a date. It's not sufficient to say that the flyer says so because of how timezones are handled, since some dates will allow 1 day earlier. Note that dates they hand out the code are not the same as dates that the code is redeemable. We need the latter.

Here's an example of another event date being added and discussion around it: https://github.com/kwsch/PKHeX/pull/4340

bdawg1989:
Ah, I see now.  Thank you for your explanation.  I will close this PR until we have that information tomorrow.

bdawg1989:
> Your dates are still incorrect. I suggest you explain your reasoning for why you decided on a date. It's not sufficient to say that the flyer says so because of how timezones are handled, since some dates will allow 1 day earlier. Note that dates they hand out the code are not the same as dates that the code is redeemable. We need the latter.
> 
> Here's an example of another event date being added and discussion around it: #4340

Taking a closer look at the image that Serebii posted, is it not sufficient to say this says "System code input deadline until June 30, 2025 (Mon)"?

image

So would I still be incorrect in assuming the date?

Thanks,

bdawg1989:
Yes, and when I translated it yesterday, I missed the part shown above where it says when it can be redeemed by.  I also accounted for the timezones the same way the rest of the dates are added in the `EncounterServerDate` class.  

And yes, the opening time to begin serial code collection is 10am their time (4/25) GMT +9 hours.  So a beginning date of 4/24 accounting for all other time zones should be correct.

Please correct me if I'm wrong.

9Bitdo:
The Lotte World Park in South Korea, Seoul open at 1000 KST(UTC +9) thus the earliest redemption for Ditto's serial code would be possible on the 24-04-2025 for countries in the (UTC-11)'s time zone. The serial code must be redeemed by 30-06-2025 KST(UTC +9), this would allow countries from (UTC +10) time zone and onwards to still be able to redeem the serial code on 01-07-2025.

i would agreed with bdawg1989 on the date submitted

PR #4482, opened 2025-04-30
Feature: Localization of Battle Templates (Showdown Set)

### LINKED ISSUE #4477: French localization improvements
Hey there.
So I'm using PKHeX in French, and wanted to make a couple edits to the French localization. There are a couple language-specific issues as per the screenshot (in the preview tooltip):
- The nature's name (for ex. "Timide" for, well, "Timid") comes after the word "Nature" (so the correct phrasing would be "Nature Timide", or even better "Nature : Timide").
- This is more of a nitpick but French adds a space between words and colons, exclamation, question marks (so for ex. instead of "IVs:", it'd say "IVs :"). Like I mentioned, it's less of an issue, since I could theoretically embed the space inside the translation string itself. 
I'll gladly work on an improvement once I know what to do for the nature string.

Cheers.

Image

kwsch:
Fixes #4477 

# Overview

This PR introduces support for localizing Showdown Sets into any language, enabling better international accessibility and integration.

## What is a Showdown Set?
A Showdown Set is a serialized, human-readable representation of a Pokémon's properties, popularized by [Pokémon Showdown](https://github.com/smogon/pokemon-showdown/) and battle simulators that preceded it. It encodes all relevant parameters for a battle template — such as species, form, level, EVs, IVs, nature, ability, item, and moves — into a compact text format. These sets are commonly copy-pasted between users, websites, and tools, and are a de facto standard for sharing competitive builds.

## Why localization?
Until now, Showdown Sets have only been available in English, limiting usability for non-English speakers. By enabling localization of all human-readable fields (e.g., move names, nature names, item names), this feature allows users to view and share and modify sets in their preferred language.

## Nuances of localization/tokenization
Localization goes beyond simple string substitution; it requires adapting the structure and grammar of the text to match the conventions of each target language. However, in languages with different grammatical rules — such as German or Japanese, the ordering of descriptors, and even the direction of reading may change. Some languages require rearranging adjectives or flipping noun-modifier pairs (e.g., “Timid Nature” vs. “Nature Timide”), while others rely on full-word descriptors instead of abbreviations.

Showdown Sets are essentially a list of properties, similar to [JSON](https://en.wikipedia.org/wiki/JSON). Humans and parsers break apart each row via [lexical analysis](https://en.wikipedia.org/wiki/Lexical_analysis#Tokenization), with most rows being prefixed by which property they are for.

To accomplish tokenization for each language, each line is classified into one of three groups of lexical token types:
* Left: `EVs: 0 Atk`, where `EVs: ` is a left-side token for the line's encoded value.
* Right: `Timid Nature`, where ` Nature` is a right-side token.
* Center: `Shiny: Yes`, where the entire line doesn't really store a value, but when present, represents a value.

Each language parser configuration contains the above 3 sets. `Nature` in English's configuration might be a right-side token, but in French, it is a left-side token. Additional translations can be stored in the parser configuration for things like Gender and Stat Names.

# Summary of Localization

This change preserves the core data structure of Battle Templates and their serialization format, while allowing for flexible localization during formatting/export. Multiple new API methods have been added to allow for importing with default localization (English), in addition to language-agnostic import via trying all localizations to find the best parse. 

Default settings continue to export the canonical English format, but users are able to override the setting to their language of choice.

# Other Improvements
In addition to localization, other settings have been added separately for the following use cases currently in PKHeX's GUI:
* Exporting sets to clipboard
* Hovering slots

Settings available (separate configuration sets for the above):
* Use a specific language or use the current program language.
* Use longer stat names (`Attack` instead of `Atk`), or even the popular [Japanese HABCDS stat labeling](https://pokebook.jp/knowledge/habcds).
* Create a custom token order, to configure which lines get shown in a set. Hide rows or rearrange the output! Note: some extra token types are included like `Awakening` and `Ganbaru`; these are not stored in Battle Templates so they will not be exported unless handled manually (hover preview)

# Future
* Adds support for Showdown's proposed changes to token syntax for EVs-Nature and Ability-HeldItem. Lines will be parsed, and a custom configuration can be specified to match (if you want to test their output format). Refer to `BattleTemplateConfig.ShowdownNew` for token order.
* Adds support for parsing directional moveslot used by Pokémon Legends: ZA. Refer to [prerelease screenshots](https://www.serebii.net/legendsz-a/pics.shtml) where moves are shown in a diamond configuration. Assumedly counter clockwise -- up,left,down,right; TBD for actual `PA9` stored move index order, and currently is not output regardless of user settings. I only intend to have it output when Context is Gen9a (or future? who knows??)


kwsch:
I'll revise HABCDS and stat read/write so that it can support `Stat:`Value syntax and be dependent on the display type, before this is merged.

PR #4499, opened 2025-05-21
Update EncounterDist9.cs

bdawg1989:
The class is already providing implementations for the interface members (like Location, EggLocation, Version, LevelMin, LevelMax), but wasn't actually declaring that it implements the interfaces.


kwsch:
It's already provided via `ITeraRaid9` which inherits from other interfaces.

bdawg1989:
> It's already provided via `ITeraRaid9` which inherits from other interfaces.

Why am I getting these errors then?

image

bdawg1989:
Nevermind.  Closing out of Visual Studio and Opening it back up fixed it -.-

PR #4512, opened 2025-06-15
DP - Add Spear Pillar related flags

Mimigris:
Explanation of the various flags:
0350 - Set ON after beating/catching Dialga/Palkia in Spear Pillar during the story. When ON, the man blocking access to Route 222 due to a blackout at Sunyshore City is no longer there, and the entry on Palkia/Dialga can be read at Celestic Town.
0351 - Set ON after catching Dialga/Palkia. Dialga/Palkia will not reappear after getting to the Hall of Fame.
0482 - Set ON after beating/catching Dialga/Palkia, set OFF after making it respawn by getting to the Hall of Fame without having caught it. The Dialga/Palkia NPC at the top of Spear Pillar will no longer be present.
1023 - Self-explanatory.
2433 - Set ON after beating/catching Dialga/Palkia during the story. Unlock the 2nd level of rematch against trainers. Applies to Platinum as well, so I added it.

I haven't looked into the Platinum flags since it was out of my scope (see the flag research section), and that it would require to resume from scratch the search on this since the Spear Pillar story is completely different there, so I just checked quickly those that could be easy to check, which ended up being the rematch flag.

Misc flags that could be added if wanted:
0685 - Set ON after beating/catching Dialga/Palkia during the story. The Galaxy grunts will be gone from Mt. Coronet.
0486 - Set ON after beating/catching Dialga/Palkia during the story. The Galaxy grunts will be gone from Spear Pillar.
0487 - Same as 0486, but for the Commanders.
0488 - Same as 0486, but for Cyrus.
0492 - Same as 0486, but for Uxie.
0493 - Same as 0486, but for Mesprit.
0494 - Same as 0486, but for Azelf.
I added none of them due to being heavily tied to the story (cutscene stuff), and making one of those disappear outside of the Mount Coronet grunts while not having finished the Spear Pillar script will lead to a crash (same if 0484 or 0485 is ON while doing the story at Spear Pillar).

Now, why did I searched those flags:
Diamond and Pearl have a glitch known as Tweaking Trick, allowing to do a lot of things, like getting to legendaries/mythical earlier without needing the key item, unlocking Fly locations earlier, and skipping some story events. Among those possibilities is getting to the Hall of Fame at 2 badges (https://www.youtube.com/watch?v=SlU_p1FEjb0). While it works and can be interesting to do for players, granting the player with a star, allowing to catch the lake guardians without having visited the Spear Pillar and done the story, allowing the National Dex to be unlocked earlier and Battle Frontier to be visited earlier, it comes with the flaw of breaking the story elements on the top of Spear Pillar. This leads to 1) some Galaxy members like Cyrus not being able to be challenged at Spear Pillar, being frozen in place and not doing anything, 2) Galaxy members in Mount Coronet and Spear Pillar not being able to leave since the Spear Pillar event cannot be completed, 3) prevent the blackout at Sunyshore City to finish, preventing the access to Route 222 without doing additional glitches or save editing, and 4) rematches being locked to those after beating Fantina since the level of rematch unlocked by beating the Spear Pillar content cannot be unlocked in this glitched state. The documentation of those flags allow players having done this bug to avoid all those issues without having to start a new save file, either by setting the Dialga/Palkia variable to 0, setting the flag 482 to OFF (only ON if you caught Dialga/Palkia), and then completing the Spear Pillar normally, or by setting the flags 0350, 0482, 0486~0488, 0492~0494, 0685, and 2433 to ON while setting the Dialga/Palkia variable to 4.

Miscellaneous changes:
- Fix Accessible being written as Accessable
- Fix Jennifer being written as Jennnifer
- Fix the Korean translation not being up to date on the lines in the Platinum flags

Questions about this pull request, now:
- Is it all good to add those flags? (no idea what is a flag that should be added or not) Should I also add the flags mentioned in this comment, or are they not considered important?
- Is there a proper order, and proper letter to put for the flags? I had no real idea what was the formatting, so I tried to do what I could but it's a bit weird.

Also added flags tied to the Slowpoke Well event in HGSS:
The Slowpoke Well flags are once more added for players having exploited Tweaking Trick during their playthrough, leading to issues later on and needing save editing (tweak in question: https://youtu.be/KK4GScMdcCc).
If a player exploited it during their playthrough, they will not be able to access Slowpoke Well without doing glitches/save editing, and Kurt's granddaughter will continue speaking as if the Slowpoke weren't rescued.
For those affected, setting the flags 0119 and 0415 ON should be all good.

Documentation about those flags:
They are all set ON when interacting with Kurt for the first time in his house.
0119 - When ON, Kurt's Granddaughter text will change, and she will worry about Kurt going to the Slowpoke Well instead of being worried about the Slowpoke. If this flag is OFF after finishing the Slowpoke Well quest, she will continue using her worried about Slowpoke line instead of her normal text.
0414 - When ON, Kurt will no longer be in his house. Set OFF when completing the Slowpoke Well quest.
0415 - When ON, the Rocket grunt blocking access to the Slowpoke Well will be gone.


kwsch:
Event flag type (used for filtering out less important flags): https://github.com/kwsch/PKHeX/blob/30d51fc8807b0813cc93107cd5a3b74ca978d5b7/PKHeX.Core/Editing/Saves/Editors/EventOld/NamedEventType.cs

We don't necessarily need every single event flag documented, mainly just ones people would find interesting to toggle.

Order: Not really specific, just try to lump together near similar flags. Usually this is numerical, but doesn't really matter.

If there is a specific pattern of flag toggling to facilitate a complete event reset (similar to what is done for BD/SP events), then an event reset button can be added to the Misc editor or somewhere else so that all it takes is a button click to do multiple flag/work changes.

Mimigris:
> Event flag type (used for filtering out less important flags): https://github.com/kwsch/PKHeX/blob/30d51fc8807b0813cc93107cd5a3b74ca978d5b7/PKHeX.Core/Editing/Saves/Editors/EventOld/NamedEventType.cs
>
> We don't necessarily need every single event flag documented, mainly just ones people would find interesting to toggle.
> 
> Order: Not really specific, just try to lump together near similar flags. Usually this is numerical, but doesn't really matter.
Understood, I adjusted some of the type flags set

Understood, I adjusted the type set for some of the flags. The order should mostly be good, so I didn't edited it. I understood that not all of the flags were worth adding (some plugins can already help you for more minor things like individual items), but since those are story flags where there could be an interest depending on what the player did, I thought they would be interesting to add. I may try to add more flags tied to similar issues, or indirectly locking the players, if I get time and motivation on documenting all of that.

> If there is a specific pattern of flag toggling to facilitate a complete event reset (similar to what is done for BD/SP events), then an event reset button can be added to the Misc editor or somewhere else so that all it takes is a button click to do multiple flag/work changes.

Having a proper interface for a collection of flags could be interesting, but I would need to take a look at how it works, and this current solution works fine for now. I cannot really check how it works for BD/SP (requires to edit an existing save, and I don't have one on me), though I suppose if it's like the Misc editors in other games it should be all good. I think adding a proper menu for the decorations in GSC could be interesting for easyness of use, so I suppose this could be attempted to be added as a practice.
On this topic, is there any way to search for flags and constants in the first generation? There is no menu for that, and I wanted to take a look at constants and flags for the Old Man Trick (iirc, it is possible to have an encounter be stored in the save file for a specific location that would be triggered when reaching the map, which can be problematic for an user too if it were to occur on an invalid mon and that they are unaware of this).

As for this pull request, I consider it to be finished, so I'm opening it 👍 .

PR #4520, opened 2025-07-06
Minor Clean

bdawg1989:
Removed excessive spacing


Lusamine:
This was intentional to line everything up so it is easier to read.
image

PR #4535, opened 2025-08-05
Update SAV_PokedexORAS.cs

hexbyt3:
Remove unnecessary return statement from SetEntry method


kwsch:
Falls under hygiene preference; put local functions at the bottom, indicate that the method returns, and the rest is ignored.

https://www.jetbrains.com/help/resharper/SeparateLocalFunctionsWithJumpStatement.html

PR #4536, opened 2025-08-05
Enhancement: Add timestamped backup rotation system

hexbyt3:
Enhances the backup system to maintain multiple timestamped backups instead of a single .bak file. This provides users with better recovery options and version history.

Changes:
- Create timestamped backups with format: filename.bak.yyyyMMdd_HHmmss
- Implement automatic rotation to keep only the most recent N backups (default: 5)
- Add MaxBackupCount setting to allow users to configure retention limit
- Apply rotation to both PKM files and save files
- Include fallback to simple .bak files if timestamped creation fails

This improvement gives users peace of mind when editing their Pokémon, knowing they have multiple recovery points available if needed.

### DIFF
--- a/PKHeX.WinForms/MainWindow/Main.cs
+++ b/PKHeX.WinForms/MainWindow/Main.cs
@@ -918,9 +918,12 @@ public partial class Main : Form
             return false;
 
         var meta = sav.Metadata;
-        var backupName = meta.GetBackupFileName(dir);
-        if (File.Exists(backupName))
-            return false; // Already backed up.
+        
+        // Create timestamped backup filename
+        var timestamp = DateTime.Now.ToString("yyyyMMdd_HHmmss");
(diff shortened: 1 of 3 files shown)


kwsch:
Not sure what the value added in this is besides some save file types (like battle revolution) not having a specific playtime to encode in the backup name. 

By default, backups are already saved with a timestamp, either the playtime or lastsaveddate/time (in-game, from console). Keeping multiple backups of the same file, if opened multiple times, is unnecessary. For save files without innate timestamps, a checksum is computed to make it a unique file name based on the data.

As it currently stands, the backup filenames already being unique via their own timestamp/checksum doesn't work with this proposed change.

hexbyt3:
I debated adding it, but a user told me that a feature like that would be useful to them so I added it to my fork.  Didn't know if you could use it or not, but I get what you mean about the added value part.  Thanks for looking anyway!

PR #4542, opened 2025-08-13
French - Update the legality file

Mimigris:
Fix a lot of weird translations/inaccuracies, translate untranslated lines.

Notes:
- BallEggCherish & BallEggMaster were missing the mention of normal, which is incorrect (some eggs were distributed in a Cherish Ball in 2017)
- EncStaticPIDShiny was using the term shiny-locked to describe it as if the error could only apply to non-shiny Pokémon set to be shiny, but from the English formulation I understand it as implying the opposite too (e.g. a shiny only Pokémon set to be non-shiny could display the warning), so I renamed it
- FatefulGiftMissing was mentioning the Mystery Gift DB being edited which wasn't the case in the original, edited
- G2OTGender: mention Pokémon Crystal instead of just Crystal since I struggled to understand Crystal meant the game rather than the name
- MemoryArgBadItem_H1 was specifying that the Pokémon cannot have held any item rather than a specific item, fixed

Others:
- No clue if the Mood and Spirit stats have an official French name, so a direct translation from the English term was used
- I translated HT (Handling Trainer) as Der. Dres. (Dernier Dresseur, Last Trainer), takes more space but clearer than using DD, or just an abbreviation that wouldn't mean anything out of the box
- TransferMet: no clue what Crown is supposed to mean, I changed it to be more accurate (from what I understand, it's supposed to be for Relocator Pokémon, so legendary beasts and Celebi, so I specified the expected potential encounter locations for them). If this is not correct, feel free to change it or tell me what it's supposed to be.


Lusamine:
> TransferMet: no clue what Crown is supposed to mean, I changed it to be more accurate (from what I understand, it's supposed to be for Relocator Pokémon, so legendary beasts and Celebi, so I specified the expected potential encounter locations for them). If this is not correct, feel free to change it or tell me what it's supposed to be.

"Crown" is supposed to reference the "Crown Beasts" being transferred up via Relocator, but this may not translate well into other languages. I don't think the original string is very clear, but I don't think you should end up with French as the only translation that is different either. I would suggest changing the English string (and yet-to-be-translated strings) to something like "Invalid Met Location, expected Poké Transfer or Relocator (Zoroark) transfer".

Mimigris:
> "Crown" is supposed to reference the "Crown Beasts" being transferred up via Relocator, but this may not translate well into other languages.

I wasn't aware this was a term used for them, the more you know.

> I don't think the original string is very clear, but I don't think you should end up with French as the only translation that is different either. I would suggest changing the English string (and yet-to-be-translated strings) to something like "Invalid Met Location, expected Poké Transfer or Relocator (Zoroark) transfer".

I totally agree on this, but if you asked me, rather than just changing the text, the message and check should be changed to separate the Crown Beasts and Celebi (would make it even clearer, and not have a message mentioning another species out of nowhere): I feel like just changing the text here would be a work half-way done, even if still a step forward, and I wouldn't know how to properly separate the checks to do something properly without help.

PR #4550, opened 2025-08-31
Update the French translation

Mimigris:
Does the following changes to the French translation:
- Translate setparse in French
- Translate lines that were untranslated in legality_fr, flags_bw_fr, flags_oras_fr
- Translate some lines in the global lang_fr file (Pass Powers, O-Powers, Gen 7 Throw Styles)
- Misc edits in encounter_fr, movesource_fr, messagestrings_fr (check commits for more details)

Some minor fixes were also applied to other flags_oras files, tix a typo where Fallarbor Town was written as Fallabor Town and Verdanturf Town as Vendanturf Town.

Note 1:
I added to some of the flags lines more details in French (e.g. specifying which Pokémon is traded with which for in-game trades, giving the class and location of trainer rematches), if it's wanted I can port them to English as well (the same would go for the Zorua line mentioned in my previous pr by Lusamine).

Note 2:
Was wondering, how is the legality file supposed to work/for what? Checking the legality of a Pokémon seems untranslated no matter which language I select while it should be translated (e.g. Japanese still displays in English), so I was wondering if I was looking into the wrong place, or if the lines were not translated for the display due to a bug.


kwsch:
Note 1: sure
Note 2: fixed locally, will push & hotfix the uploaded exe once the new raids update in ~2 hours

PR #4565, opened 2025-09-25
Added a couple of translations in French and Chinese

XieonCrypto:
Fixed a few lines in French and Chinese language files that were still in English, and put them in the correct language


Lusamine:
Untranslated lines are not necessarily "errors" like you said in the title of this PR. Often, translateables are added later and it simply means nobody has added a translation yet.

PR #4571, opened 2025-10-11
Update WB8.cs

hexbyt3:
Manaphy eggs have MetLocation = Default8bNone (65535), which requires MetDate to be 0/0/0 according to the legality validator. Previously, the MetDate was being set to the distribution window date for all Pokémon, causing eggs to fail validation.

Changes:
- Clear MetDate components (Year/Month/Day) for Manaphy eggs after setting MetLocation to Default8bNone
- Skip setting MetDate for Manaphy eggs in the date assignment logic

This fixes the "Met Date is not a valid calendar date" error for Manaphy egg distributions (CardID 1701) in SysBot.NET projects.

### DIFF
--- a/PKHeX.Core/MysteryGifts/WB8.cs
+++ b/PKHeX.Core/MysteryGifts/WB8.cs
@@ -548,7 +548,8 @@ public sealed class WB8 : DataMysteryGift,
     private void SetEggMetData(PB8 pk)
     {
         pk.IsEgg = true;
-        pk.EggMetDate = EncounterDate.GetDateSwitch();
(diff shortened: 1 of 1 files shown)


kwsch:
Wouldn't a Manaphy egg (that is still an egg) have an EggDate?

hexbyt3:
> Wouldn't a Manaphy egg (that is still an egg) have an EggDate?

Damn daddy, you're right.  I went with the random dates in between the valid dates [Serebii has listed](https://www.serebii.net/events/dex/490.shtml).

PR #4573, opened 2025-10-12
Update Event Flags and translations

Ka-n00b:
I started this last weekend. Planned to push this a few days ago, but didn't get the chance to.

* Added new event flags and constants
* Fixed and renames some existing event flags and constants
* Added translations to many (not all) flags/constants and message strings


Lusamine:
How did you translate all of these languages?

Ka-n00b:
> How did you translate all of these languages?

Sorry, I didn't notice your post until now.

A lot of translations in the event flag sections were done for simple statements like "Received," "Completed," and "Disappeared" by referencing other translated lists in those same languages. I also looked up Wikis, Let's Plays, and checked in-game myself to a point in those specific languages to find out the names for any nouns that I couldn't find in those languages on Bulbapedia.

As for many concise sentences, I did consult ChatGPT to find out what they are while making sure that any official nouns were retained if they were present.
For Traditional Chinese translations, I took the existing Simplified ones and converted their specific characters.

PR #4575, opened 2025-10-14
Update flags_bw_fr.txt

Mimigris:
Correction of a very minor typo brought in #4573 , and adjustement of the Abyssal Ruins level to match with how it is called on the wiki


Lusamine:
Thanks, we figured any discrepancies would be fixed by native speakers and it's probably better than untranslated.

Mimigris:
> Thanks, we figured any discrepancies would be fixed by native speakers and it's probably better than untranslated.

Since the translations reuse/transform existing lines to match with the languages, the translations end up being most often accurate so it's not an issue (I only had to adjust two lines over minor details, so yeah I too think it was the best option here, don't worry 👍 ).

PR #4586, opened 2025-10-23
Misc. Changes

hexbyt3:
1. Expose Internal Data (internal → public)
 - For plugin use (SV/SWSH Seed Finder, Export Data, etc.)

2. Dark Mode Enhancements
 - Program.cs: Added design-time check to prevent Settings null reference crashes
 - UI Controls: Dark mode color fixes for better text visibility:
   - PKMEditor: Dark theme for tabs and backgrounds
   - StatEditor: Dynamic foreground colors based on background brightness
   - BoxEditor: Dark background, suppressed wallpaper in dark mode
   - SAVEditor & VerticalTabControl: Similar dark mode theming

### DIFF
--- a/PKHeX.Core/Legality/Encounters/Data/Gen7/Encounters7GG.cs
+++ b/PKHeX.Core/Legality/Encounters/Data/Gen7/Encounters7GG.cs
@@ -3,12 +3,12 @@ using static PKHeX.Core.GameVersion;
 
 namespace PKHeX.Core;
 
-internal static class Encounters7GG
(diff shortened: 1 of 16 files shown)


Lusamine:
Ideally, changes that are unrelated should be separate pull requests, not just combined into "misc changes". This makes it easier to make changes if one proposal is accepted and another is not, and it keeps the repo history readable when merged.

Since this was already done in this way, you should wait on a comment for whether either of these changes is wanted before you do anything.

PR #4587, opened 2025-10-23
Expose encounter data through read-only public API

hexbyt3:
## Summary
 Exposes encounter data classes and arrays as public with read-only access for plugin use, while maintaining zero performance overhead for internal PKHeX code.

  ## Changes
  - Make encounter data classes public (Gen 7-9: Encounters7GG, Encounters8, Encounters8Nest, Encounters8a, Encounters8b, Encounters9)
  - Internal arrays renamed with `Array` suffix (e.g., `SlotsArray`, `Encounter_SVArray`)
  - Public `IReadOnlyList<T>` properties maintain original names (e.g., `Slots`, `Encounter_SV`)
  - Internal generators updated to use array fields directly

 Prevents external plugins from accidentally or intentionally modifying PKHeX's encounter data while still allowing read access for tools like seed finders and data exporters.

### DIFF
--- a/PKHeX.Core/Legality/Encounters/Data/Gen8/Encounters8Nest.cs
+++ b/PKHeX.Core/Legality/Encounters/Data/Gen8/Encounters8Nest.cs
@@ -278,9 +278,9 @@ public static class Encounters8Nest
     private const AbilityPermission A3 = AbilityPermission.Any12;
   //private const AbilityPermission A4 = AbilityPermission.Any12H;
 
-    internal const int SharedNest = 162;
-    internal const int Watchtower = 126;
-    internal const int MaxLair = 244;
+    public const int SharedNest = 162;
+    public const int Watchtower = 126;
+    public const int MaxLair = 244;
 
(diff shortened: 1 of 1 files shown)


hexbyt3:
Sorry, accidentally deleted other unused branches and deleted the good ones in the process.

kwsch:
I've decided to not have these made as `public` for dll/etc use.

Recommendations:
1. Use the `EncounterMovesetGenerator` to fetch encounters. This has worked fine for ALM, as users really do not need direct info access besides searching for an encounter that meets their needs. The encounter generator can return sufficiently targeted search results without needing direct array access for later manual filtering.
2. If you really really really need access, then Reflection can be used to get private/internal fields from a class.
3. For species, you can obtain the current max from the save file/pkm object's property `MaxSpeciesID`.
4. For wild ball legality, using the `GetWildBalls` method returns the relevant bitflag value; no need for individual value fetch. If needing a list of legal balls for an encounter, refer to `BallApplicator`.

PR #4588, opened 2025-10-23
Fix dark mode and add design-time safety checks

hexbyt3:
Improves dark mode appearance by dynamically adjusting text colors based on background brightness. Prevents Settings-related crashes during design-time by adding proper null checks.

Changes:
- Adjust foreground colors in StatEditor when background is bright
- Apply dark theming to tab controls and editor panels
- Suppress box wallpapers in dark mode for better contrast
- Add design-time checks to prevent null reference errors in Program.cs when editing in VS Designer

### DIFF
--- a/PKHeX.WinForms/Controls/PKM Editor/PKMEditor.cs
+++ b/PKHeX.WinForms/Controls/PKM Editor/PKMEditor.cs
@@ -78,6 +78,23 @@ public sealed partial class PKMEditor : UserControl, IMainEditor
         TB_EXP.MouseWheel += WinFormsUtil.MouseWheelIncrement1;
         TB_Level.MouseWheel += WinFormsUtil.MouseWheelIncrement1;
         TB_Friendship.MouseWheel += WinFormsUtil.MouseWheelIncrement1;
+
+        Load += (_, _) =>
+        {
+            if (DesignMode || !Program.Settings.Startup.DarkMode)
+                return;
+
+            var darkBg = Color.FromArgb(30, 30, 30);
+            BackColor = Hidden_TC.BackColor = TC_Editor.BackColor = darkBg;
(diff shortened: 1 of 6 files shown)


ELY3M:
your dark mode need some work.   
see this image 
darkmode

hexbyt3:
Thank you.  From this:
image
to the image you shared above, looks like an improvement to me.  Net 10 is coming next month and should give us more options to play around with.  This was just a minimal change to help cover the spots that were not covered before.

sora10pls:
I disagree with the implementation of box wallpapers when running in dark mode.

In the many games that have unique box wallpaper designs prior to SWSH, the game will auto-populate the boxes with wallpapers in order by index (not considering special wallpapers, unlocked in post-game), in order to make each of them look unique. Beyond this, there are many people out there - myself included - that use certain box wallpapers to match the contents inside that box.

Defaulting to a null image for the sake of looking better in dark mode causes inconsistency with how it's displayed in-game, as well as goes against the points I mentioned earlier.

The only real reason I could see for wanting a null background image is for the very basic wallpaper used for SWSH/SV, which in-game is displayed as a lapse of various colors updating in real-time as you're in the boxes menu, since I would agree that the very bright color doesn't clash very well with dark mode.

hexbyt3:
That is a valid point.  I did not consider the use of wallpapers and was really just going for the overall dark-theme look completely.  I think I'm going to scrap this PR completely for now until we can see what NET10 natively supports.  Thanks!

Lusamine:
My initial gripe with this wasn't about the backgrounds, but the fact that dark mode on .NET 9 looks horrible and patchy still. I would prefer a proper implementation in .NET 10 rather than adding jank to throw out later.

PR #4601, opened 2025-10-27
Minor tweaks

CScorpion-h:
Fix SAV9ZA Save Box Data Function
Fix ZA Pokemon size display to use 5 size classes instead of 9


kwsch:
Thank you for reporting these; fixed via different implementations.

pcdata/boxdata fix:
https://github.com/kwsch/PKHeX/blob/19284fef79dc853dfe1f0fe5b35d6169ab848a8b/PKHeX.Core/Saves/SAV9ZA.cs#L168

scale label change:
https://github.com/kwsch/PKHeX/commit/b407d8f0a06b12a1438f21447df26651d6830d2c#diff-f3147dcc5f0b58832bd9fb74044066f004c1fcb8c55275e8d208be538d25d9d7R16-R165

PR #4604, opened 2025-10-27
Add Alpha status box sort option for PLA/ZA

Omni-KingZeno:


kwsch:
Scale sort already does this and was sufficient for Legends: Arceus, but I guess why not.

PR #4630, opened 2025-11-07
Refactor gender and nature setting for PA9 class

hexbyt3:
PA9 Pokemon use LumioseRNG which correlates EC/PID/IVs/Gender/Nature/Scale from a single seed. Modifying these properties after encounter generation breaks the RNG correlation.

Changes:
- Skip gender modification for PA9 to preserve RNG correlation
- Skip nature modification for PA9 to preserve RNG correlation
- Use SetPlusFlagsEncounter for PA9 instead of generic LegalityAnalysis to apply Plus Move flags correctly based on encounter data

Fixes issue where PA9 gift Pokemon like Totodile would become illegal after Showdown import.

### DIFF
--- a/PKHeX.Core/Editing/CommonEdits.cs
+++ b/PKHeX.Core/Editing/CommonEdits.cs
@@ -213,12 +213,15 @@ public static class CommonEdits
             pk.SetMarkings();
 
         pk.SetNickname(set.Nickname);
-        pk.SetSaneGender(set.Gender);
+
+        if (pk is not PA9)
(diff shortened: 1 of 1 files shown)


Lusamine:
I'm going to remind you to please do not submit ANYTHING related to legality in issues or pull requests until encounters are properly handled in the base program.

Per the release and the one before:
> Please do not report issues with this release's handling with legality/encounters. We are aware of what still needs work.

I will be closing further requests from you until we have given the go-ahead to submit issues.

PR #4692, opened 2026-01-13
Prevent Moves from Highlighting in EncounterDB when opening

RandomGuy155:


kwsch:
ty, implemented for all 3 popup forms (encdb, mgdb, pkmdb): https://github.com/kwsch/PKHeX/commit/c7b5777068463a9e4f083848d468eb86c21ab658#diff-08cc001c6c0d5099a6160e2fd220324221b119ec147e29281352a64152874c61R165-R170

I was doing tweaks on it so I included with my commit + you as contrib :D

PR #4712, opened 2026-02-08
Add a search interface for visually filtering all slots 

kwsch:
image

Notice the button in the top right corner of the box slots? Clicking it will open the search GUI, similar to the PKM Database (it's actually the same control). By entering your search criteria and clicking Search!, the program will apply a search filter. Any Pokémon slot that does not match your criteria will be made 75% transparent, to visually differentiate the matches.

In the Advanced tab you can use the same batch editor filters for precise property filtering, same as the Batch Editor or PKM Database.

image

In addition, the following shortcut key-combos are provided:
* Alt-Click: Clears the current search.
* Shift-Click: Jump to the next box with a result.

Feel free to discuss in Discord; might make a few more visual fixes before this is merged (soon).

image


Lusamine:
Wonder if it is worth slightly reducing the transparency for non-matches? Some of the Pokémon are getting difficult to see against backgrounds.

Original 25%:
image

35%:
image

50%:
image

PR #4714, opened 2026-02-10
implement missing `StreetName` property in SAV9ZA.cs

realgarit:
---
name: This PR implements the `StreetName` property in SAV9ZA.cs
about: The block constants for `KStreetName` were already defined in SaveBlockAccessor9ZA.cs, but the property was not exposed in the SAV9ZA.cs class.
title: ''
labels: enhancement
assignees: ''
---


kwsch:
DLC properties are not present in the base game save file, so it would throw an exception if the get/set is called. For this reason alone, it is not a consistent property to include in SAV9ZA.

This property is already surfaced in the GUI via the DLC editor, where the get/set is manually implemented:
https://github.com/kwsch/PKHeX/blob/20905cbe67d92545f80e82ac7bd0a7685401cab3/PKHeX.WinForms/Subforms/Save%20Editors/Gen9/SAV_Trainer9a.cs#L130-L140
Notably, the setter should be clearing the buffer before writing the string, rather than only overwriting the characters provided for the length of the provided (not entire) string.

Due to SCBlock save files having a ton of loose properties, implementing every single one into the SaveFile as get/set isn't really sustainable or worthwhile. Similar to how the games do it, the consumer does the read/write on the Block data rather than the savefile itself doing the work.

realgarit:
Thanks for the insight! I see your point.

PR #4735, opened 2026-02-27
Translate remaining lines of the program to French

Mimigris:
Translate remaining translatable lines of the program to French (excluding Switch flags, not interested with touching much Switch games).

Notes:
Several of the lines are a bit unclear for me. This covers:
- Main.L_Resort ("Resort", no context given about the use)
- SAV_Trainer7.L_DaysFromRefreshed ("Days from Refreshed:", not understanding what it means, but it's used in the Misc tab of the Gen 7 Trainer Info)
- LocalizedDescription.FocusBorderDeflate ("Focus border indentation for custom drawn image controls.", not understanding either what it means/entails, may need a picture to understand)

I either left them untranslated, or had to do a translation based on what I guessed: would it be possible to get an explanation of those lines?

Some lines used & as some sort of marker. I was unsure what it meant, as it is a non-printable character in those contexts and doesn't seem to do anything. I suppose it may be used to indicate to which hotkey the line is associated (with the letter after & being the hotkey), but is it supposed to be that/is the presence of it even required?

Would prefer to have those two interrogations be solved before merging, hence the draft pull request. I'll probably be a bit away this weekend so not much adjustement doable, but I should be able to edit the pull request as needed Monday.

Other change:
- Fix GO Park being listed as Go Park in some lines (fixed everywhere except changelog 2020, don't think it should be edited for historic purpose)

---

As the amount of line is quite dense, with a lot of game specific vocabulary, lines spread across a bit everywhere, and adjustements needed tied to window display, there will probably be additional pull requests in the future to adjust those lines. I may also see if I remove title casing in the events and constants French files, as it is a bit odd to see in French, and making those consistants is a bit hard. Outside of those points and revisions, though, the translation of the translatable lines should now be pretty much finished in French.


kwsch:
More lines are now available for translation :)

1. StorageSlotType uses the old non-enum translation logic, which is all the `Main.L_*`. There's some custom logic with handling the enum names and not blacklisting them from being translated, so it's kinda clunky. The Resort is the Poké Pelago https://www.serebii.net/ultrasunultramoon/pokepelago.shtml
2. Likely is days since last using Pokémon Refresh (interacting with your pokemon) https://www.serebii.net/sunmoon/pokemonrefresh.shtml
3. FocusBorderDeflate is the amount of pixels inwards to draw the dashed-gray border, similar to when programs highlight the outer edges when you press tab. Since the PictureBoxes have their edges stacked on top of each other (to get a 1px grid), we configure a -1 pixel edge of this selection border, and allow people to configure it to be closer inward to the interior with higher values, because why not?
4. The usage of Ampersand is useful for toolstrips; shows as an underlined char to allow for easy hotkeys https://learn.microsoft.com/en-us/dotnet/desktop/winforms/controls/how-to-create-access-keys

Looks like the merge conflict is because I removed one of the un-translated lines which you've now translated...

kwsch:
Regarding `1.`, I've refactored how the extra slots have their labels created, so that it's a little more transparent on where these values are from (since they're now all grouped to the enum).

Mimigris:
Just took care of the merge conflicts, as well as translating the remaining lines and new lines 👍

I'm not really sure to understand all to the ampersand thing (didn't understood if there was a way to make characters appear underlined in the program, nor if it mattered to map them), so for now I don't plan on personally touching them up more, but if you have any idea on improving it feel free to.

With this, unless there are errors or other things to take care of, this translation pull request should be ready to be merged!

By the way, not tied to this pull request, but I tested a bit the Events tab with the categories on the Nightly: outside of sorting of individual lines being needed (not unsurprising as categories weren't much used before), this all works really well!

Something I've noticed with this though, but since the non-English, Japanese and French event lines for gen 3 are using old files lacking tons of lines, all the lines not included in their languages are simply and purely missing when consulting the Events tab. Those files would need to be updated to use the same formatting than the other files ultimately.

And on another fully unrelated note, I checked a bit the recent commits tied to VC FRLG (I always find interesting to see what needs to change to support new releases), and I noticed that an item list was made to ban some items as not being legal (unavailable in FRLG without connecting to non-FRLG): if I'm not mistaken, the Premier Ball and Dive Ball are also unavailable in vanilla FRLG, so they should probably be added to that list (unavailable as held item + as caught ball). Not sure if it matters much in the end (legality check to support a VC game that may not be needed in a near future, depending on if they release VC RSE), but just thought it could be mentioned.

Mimigris:
Oh and I'm fine with working on adjusting the categories used by events and constants lines + inaccurate gen 3 lines if needed, forgot to mention that in my original comment.

kwsch:
> Oh and I'm fine with working on adjusting the categories used by events and constants lines + inaccurate gen 3 lines if needed, forgot to mention that in my original comment.

Sure; if you want extra categories (don't go too wild though) be sure to update the category parsing. Ideally we can keep it 1 char only for a category identifier

Mimigris:
> Sure; if you want extra categories (don't go too wild though) be sure to update the category parsing. Ideally we can keep it 1 char only for a category identifier

Will take a look at the lines needing editing tomorrow and see if additional categories will be needed then 👍 
From memory, perhaps a regular Items category will be needed (lot of regular items for instance in FRLG, currently classified as None), but without looking at the lines directly that's all I can think for now.

PR #4743, opened 2026-03-03
Update CHS translation

wubinwww:


kwsch:
The popup doesn't get translated properly; is something I'll have to rework the deferred initialization logic for the automatic translation scrape in `DevUtil`. Not your issue, it's mine :)

PR #4769, opened 2026-04-03
Allow ph4 decryption

Manu098vm:


kwsch:
Merge conflicts due to commit that was pushed a couple minutes after; just so happens that the same fix was added++ :)

Issue #4785, opened 2026-04-21
Some Artwork Pokemon Sprites have artefacts when rendering Sprite Glow


LegoFigure11:
**Describe the bug**
Some (not all) sprites have artefacts when rendering the sprite glow. Particularly noticeable when using dark mode, but present in both. 

**To Reproduce**
1. Load any save file that uses the Artwork Pokemon Sprites (eg. LZA)
2. Hover over (light mode) or view (dark mode) one of the affected Pokemon (eg. Charizard, Starmie, Gardevoir, or Talonflame) in the boxes or editor.

**Expected behavior**
The sprite glow should render without artefacts.

**Screenshots**

<img width="69" height="62" alt="Image" src="https://github.com/user-attachments/assets/85ee1038-057b-43b7-8565-49163e4ca404" />

<img width="93" height="60" alt="Image" src="https://github.com/user-attachments/assets/881ada9d-fe74-4f4f-bfb5-27c89186673e" />

**Additional context**
Tested on latest official release, 26.4.12+14 pipeline build from pporg, and locally built at 5bf1e2cf4. Noticed on some but not all sprites. Issue is not present in 2321e76, strongly implying that the Champions sprites added in 13110ad are to blame.

When viewing the sprite directly from https://raw.githubusercontent.com/kwsch/PKHeX/refs/heads/master/PKHeX.Drawing.PokeSprite/Resources/img/Artwork%20Pokemon%20Sprites/a_6.png, no artefacts are present. When viewing the file in the cloned repo on my local machine, affected sprites have some additional colored pixels with 0 or very low opacity which the program is generating the glow around.

<img width="1238" height="849" alt="Image" src="https://github.com/user-attachments/assets/1843fde4-ada1-499f-80b6-e52672b70c7d" />

I was initially willing to write this off as a "just my machine" quirk to do with git compressing the files or something as the raw file directly on github is unaffected, but seeing as it was also present in the pipeline builds I figured it was worth reporting.

BtEtta:
I've actually been working on corrected PNGs for this issue. Using a modified version of oxipng to treat all pixels below a given opacity threshold as fully transparent then normalising all such pixels to completely transparent black pixels.

This would also have the nice side effect improving compression rates for all PNGs (even without normalising almost-transparent pixels). I'm still testing the best threshold to change the least amount of data, but looks like a saving of around 3-4MiB on the compiled EXE.

[...]

PR #4788, opened 2026-04-23
Fix/optimisation pass on all PNG images

BtEtta:
Additional fixes for #4785

As per the the discussion there, images were passed through a version of [oxipng](https://github.com/oxipng/oxipng) where the `--alpha` filter was modified to treat all pixels with an opacity of 4 or lower as fully transparent.  
By changing a comparison in the oxipng source in `/src/reduction/alpha.rs` from
`if pixel.iter().skip(colored_bytes).all(|b| *b == 0) {`
to
`if pixel.iter().skip(colored_bytes).all(|b| *b <= 4) {`

The following command line was then used:
`oxipng-mod.exe --opt 4 --recursive --preserve --strip safe --alpha --nb --nc --np --ng <directory>`

`--opt 4` from personal experience represents the last optimisation level before extra savings realised come at excessive processing time  
`--recursive` includes files in subdirectories  
`--preserve` preserves existing file attributes/dates  
`--strip safe` removes unneeded PNG chunks — although I'm unsure if any files had chunks that would be removed  
`--alpha` performs the optimisation on the alpha channel, by default it would set all fully transparent pixels to black. In the modified version it uses a threshold of opacity <= 4  
`--nb --nc --np --ng` instruct oxipng not to change bit-depth, colour type, colour palette, or make the image greyscale respectively — as mentioned in the issue previous attempts at optimisation have caused problems by allowing such changes to occur

Original file size: 21,748,029 bytes (20.74 MiB)
Optimised file size: 17,015,413 bytes (16.23 MiB) — this is slightly higher than the results in the issue due to the use of the `--nb --nc --np --ng` options


BtEtta:
Forgot to mention that I did check for artifacts observed from previous optimisation attempts in both light and dark mode across all sprite display types and found no issues. Nor could I observe any problems with any GUI images, notably the menu icons that invert colour when in dark mode do so correctly.

PR #4791, opened 2026-04-25
Add Mirage Island editing for RS/E saves

joelsrzalvarez:
## Motivation
While hunting shiny Latios through Mirage Island, I found it inconvenient to rely on AR codes or external plugins just to control the Mirage Island match value. This change adds native support for that workflow directly in PKHeX for Gen 3 Hoenn saves.

## How it works
When triggered from the Misc Editor, the save reads the PID of the first party Pokémon and stores its low 16 bits as the Mirage Island match value. This mirrors the in-game comparison used by Ruby, Sapphire, and Emerald saves.

## Changes
- add `ISaveBlock3MirageIsland` for shared Gen 3 Hoenn save support
- add Mirage Island value accessors to `SaveBlock3LargeE` and `SaveBlock3LargeRS`
- add shared Mirage Island helpers to `SAV3`
- add a Misc Editor UI action to set the Mirage Island match value from the first party Pokémon PID
- hide the Mirage Island UI for saves that do not support it, such as FR/LG
- add core tests for:
  - RS/E Mirage Island value read/write
  - empty party handling
  - setting the value from the low 16 bits of the first party Pokémon PID

## Testing
- added core tests for `SaveBlock3LargeE` Mirage Island value read/write
- added core tests for `SaveBlock3LargeRS` Mirage Island value read/write
- added core tests verifying `TrySetMirageIslandFromFirstPartySlot()` returns `false` when the party is empty
- added core tests verifying `TrySetMirageIslandFromFirstPartySlot()` stores the low 16 bits of the first party Pokémon PID for both Emerald and RS
- ran `dotnet test PKHeX.sln`

### Tests
Tests

### Mirage Island appears
mirageisland

### Tab is not showing in LG/RF save file.
MirageLGFR

### Tab in R/S/E save files.
MirageRSE

### DIFF
--- a/PKHeX.Core/Saves/Blocks/Gen3/SaveBlock3LargeE.cs
+++ b/PKHeX.Core/Saves/Blocks/Gen3/SaveBlock3LargeE.cs
@@ -1,10 +1,11 @@
+using PKHeX.Core.Saves.Substructures.Gen3;
 using System;
 using System.Collections.Generic;
 using static System.Buffers.Binary.BinaryPrimitives;
 
 namespace PKHeX.Core;
 
-public sealed record SaveBlock3LargeE(Memory<byte> Raw) : ISaveBlock3LargeExpansion, ISaveBlock3LargeHoenn, IRecordStatStorage<RecID3Emerald, uint>
+public sealed record SaveBlock3LargeE(Memory<byte> Raw) : ISaveBlock3LargeExpansion, ISaveBlock3LargeHoenn, IRecordStatStorage<RecID3Emerald, uint>, ISaveBlock3MirageIsland
 {
     public Span<byte> Data => Raw.Span;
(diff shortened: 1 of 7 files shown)


kwsch:
Closing; I decided it wasn't worth a dedicated tab for Mirage Island, and a separate interface/offset interaction is the wrong pattern for interacting with this value.

There is already a plugin which does this functionality and more, so I don't feel it is necessary to have more than a simple cheat that does the first party PID copy. Existing plugin:
https://github.com/Bl4ckSh4rk/PKHeXMirageIslandPlugin

All we need to do is get the u16 at offset 0 from the Party Data span, and write it to the corresponding Work value record which the Mirage Island check uses (0x24).

The "Other" tab can be a catch-all for misc cheats, and can remove buttons if the game does not support them. No buttons remaining => remove tab.

PR #4830, opened 2026-07-02
Localize Settings/About/Block Data editors; decode Gen3 Chinese fan-translation text

doomsday616:
## Summary

This PR extends the existing `lang_*.txt` translation system to GUI areas that were previously English-only, and adds optional decoding of Chinese fan-translation strings in Gen3 saves.

### Settings editor localization
- New `PropertyGridLocalization` helper wraps each settings section with a `TypeDescriptor` provider so the PropertyGrid can localize tab names, property display names, categories, bool/enum values, dropdown standard values, and expandable object summaries via the existing lang resources.
- Values render translated but write back correctly: translated enum/bool text maps back to the original value; ambiguous translations defer to the default converter.
- Providers are registered once per settings instance (`ConditionalWeakTable`) so reopening Settings does not stack providers; the localizer dictionary is swapped on language change.
- `GameVersion` dropdown values are localized in all 10 languages by reusing each language''s own game-name strings; combined versions (e.g. RSE, FRLG) are hand-translated for zh and use readable English elsewhere.

### SAV tab Block Data localization
- All three block editors now route their PropertyGrid through the same localization: the Simple Editor fallback form (Gen1-4, with a translated title), `SAV_Accessor` (Gen5-7), and `SAV_BlockDump8` (Gen8+).

### Options -> Language submenu
- Replaces the `ToolStripComboBox` inside the Language menu with a regular drop-down submenu of checkable items.

### Localized About shortcuts
- The Shortcuts tab loads a `shortcuts_{lang}` resource matching the current program language, falling back to English. Adds `shortcuts_*.txt` for the 9 non-English languages, and updates the English hint to the current Unicode setting path.

### Gen3 Chinese fan-translation text decoding
Chinese fan translations of the Gen3 games encode Chinese characters as two-byte sequences (lead `0x01-0x1D` + second `0x00-0xF6`) indexing the 6763 GB2312 hanzi in code-point order in ranges of 247, with lead `0x06` reserved. `StringConverter3Zh` implements the codec; `StringConverter3` uses it for non-Japanese strings when active.

- Controlled by a new setting (SaveLanguage -> `Gen3ChineseFanText`): **Auto** (default) enables it only when a loaded Gen3 save looks fan-translated (box names begin with two-byte sequences; two hits required so official European saves with accented box names cannot false-match), plus Always/Never overrides. Official saves are unaffected.
- Encoding round-trips: translated names written back produce byte-identical sequences to what the game itself writes (verified against in-game data, including characters adjacent to the reserved range).
- While active, the East-Asian nickname character flag and Gen3 trash-byte rules are skipped for Gen3 entities, since those buffers legitimately contain the fan encoding.

## Testing
- Built with 0 warnings / 0 errors.
- Settings tabs, property names, categories, enum dropdowns, bool values, and Block Data editors verified localized under zh-Hans/zh-Hant and unchanged under en; enum/bool edits write back correct underlying values; language switching round-trips without stale dictionaries.
- Fan-translation decode verified against a real fan-translated FireRed save: box names and species nicknames decode correctly, English names are unaffected, and re-encoding produces byte-identical output to the game''s own encoder.


Lusamine:
Your previous pull request was better because it only handled one fix: adding the ability to localize places that were not localized. Good practice should separate features by PR, not have a giant PR with multiple changes added at once, which makes it harder to review and troubleshoot.

I will also let you know at this point that PKHeX does not intend to provide romhack support, which includes gen 3 Chinese romhack text translations. If you had opened two separate PRs, one to localize, the other for the Chinese fan text, then they could be reviewed separately.

doomsday616:
Closing per review feedback - splitting into two focused PRs: one for localization coverage only, one (separate) for the Gen3 Chinese fan-text decoding. Thanks for the guidance.

kwsch:
Gen3 fanmade Chinese encoding won't be accepted. Save your tokens :)

PKHeX does not support ROM Hacks as there are way too many permutations of what COULD be supported and the extreme hassle of how they complicate and interweave with mainline features. They can always be supported via plugins with custom PK* and SaveFile implementations (via SaveUtil.CustomSaveReaders). Separate maintenance, drop-in support as needed.

doomsday616:
Understood, thanks for the feedback. Split as suggested: localization-only changes are back on #4829 (reopened, now also covering the previously missed spots), and the Gen3 Chinese fan-text decoding is submitted separately so they can be reviewed independently.

PR #4831, opened 2026-07-03
Add opt-in Gen3 Chinese fan-translation text decoding (auto-detected)

doomsday616:
## Summary

Split out from #4830 per review feedback, so it can be evaluated independently of the localization work in #4829.

I understand PKHeX does not intend to provide romhack support, so feel free to decline this one - submitting it separately as suggested so the decision is explicit and the localization PR is unaffected.

### What it does

Chinese fan translations of the Gen3 games encode Chinese characters as two-byte sequences: a lead byte (0x01-0x1D) followed by a second byte (0x00-0xF6), indexing the 6763 GB2312 hanzi in code-point order in ranges of 247 (lead 0x06 is reserved/skipped by the font). The mapping is algorithmic - no lookup table file is required. Verified against a real save: box names and all unnicknamed species names decode correctly and encode round-trips.

- `StringConverter3Zh`: the GB2312 table plus the two-byte codec, hooked into `StringConverter3` Load/Set for non-Japanese strings only.
- Setting `SaveLanguage.Gen3ChineseFanText` (Auto/Always/Never, default Auto). Auto enables decoding only when a loaded Gen3 save looks fan-translated (two or more box names starting with a two-byte sequence; a single accented character in an official European save cannot false-match). Official saves are unaffected: with the feature inactive, behavior is byte-for-byte identical to master.
- While active, the East-Asian nickname character flag and Gen3 trash-byte rules are skipped for Gen3 entities, since those buffers legitimately contain the fan encoding.

## Testing
- Built with 0 warnings / 0 errors; test suite passes.
- Real fan-translated FireRed save: box names (盒子1-14) and species nicknames (波波, 小火龙, 小拉达, 绿毛虫, 杰尼龟) decode correctly; Chinese/mixed nickname edits round-trip; English names (PIKACHU) unaffected.
- Official saves: auto-detect returns false; converters behave identically to master.


Lusamine:
https://github.com/kwsch/PKHeX/pull/4830#issuecomment-4871990126

We were telling you not to submit this at all because it will not be accepted. The program only supports official games, not ROM hacks. 

In the future, if you want to submit multiple major changes, you should talk to us about whether it is wanted (GitHub or Discord) and submit them as different PRs if so.

doomsday616:
Sorry, my bad - I read the earlier comment as 'split them so they can be reviewed separately' rather than 'don't submit the romhack one at all'. I'll keep that part in my own fork. Will ask first on Discord next time before opening anything major. Thanks for the patience.

Lusamine:
Oh no, I meant that good code submission practice in general should separate unrelated fixes by pull request. The more unrelated changes you make in one PR, the harder it is to review it. What if one feature is unwanted but another one is? In this case, it allows the developer to choose to reject one feature and keep the other feature for review.

doomsday616:
Ah, that makes sense - thanks for spelling it out. That's actually a good general rule I should've just followed by default rather than needing it pointed out. Appreciate you taking the time to explain rather than just closing it without a word. Will keep unrelated changes in separate PRs going forward, and check in first for anything bigger like this one.

PR #4834, opened 2026-07-04
Localize save editor PropertyGrid fields

doomsday616:
## Summary
- Localize additional save-editor PropertyGrid surfaces using the existing PropertyGridLocalization helper.
- Add translated property, category, enum, and description keys for supported languages.
- Extend translation generation so these save-editor PropertyGrid keys remain stable when translations are regenerated.

## Scope
This PR is limited to PropertyGrid display localization for existing save-editor data surfaces. It does not include Gen3 Chinese fan-translation string decoding or any save-string behavior changes.

## Validation
- `git diff --check origin/master..HEAD`
- `dotnet build PKHeX.sln -c Release --no-restore`
- `dotnet test PKHeX.sln -c Release --no-build --no-restore`


Lusamine:
Is there a reason you opened both of these pull requests to localize something that was already rejected? Your merged pull request had edits with an explanation of why PropertyGrid localization .

https://github.com/kwsch/PKHeX/commit/b9d85427cba2a875b0bb103edfb0be4d3e39e6a3

> Am going to strictly refuse PropertyGrid localization of SaveFile/Accessor property grid displays. There are far too many properties to localize, and the property names are subject to change as the codebase is maintained. Settings are much more "stable" and thus acceptable for property name localization.

> Downside: the shortcuts text file was very tedious to update every time I added a new hidden shortcut. A better path forward would be to use attributes on methods to tag them as "hey a shortcut exists here, the hotkeys trigger `behavior` string", and the program dumps them into the shortcut log.

> The WinFormsTranslator probably needs to be rewritten to be an object (less primitive obsession) so that it can store multiple kvp sets for a language, and dump them to separate files with specific serializer methods. The GUI localization for form controls can be one, shortcuts another, property grid another. The property grid localization can be better arranged so that LocalizedDescription is alphabetically logged below the PropertyName, and can also have PropertyType if it is class-based.

> Too much effort for very little value add, at least for me.

Am I misunderstanding and you are changing something different?

PR #4848, opened 2026-08-05
Exclude same gender rival in PokeGear callers for HGSS

ScarlettHaze:
Added functionality to exclude the same gender rival from PokeGear phone list.
Also removed Bike Shop since you can't legally get that number saved in the game.


sora10pls:
Shouldn’t the same-gender character be excluded? Not the opposite-gender?

ScarlettHaze:
lmao you are right I had a brain fart. Fixing it now!

PR #4851, opened 2026-08-07
fix: stop nickname/OT-name corruption on unrelated edits

doctorllll:
## Summary
- Editing any field on a Pokémon (level, moves, ball, etc.) and clicking Set unconditionally re-encoded `Nickname`/`OriginalTrainerName` from the ViewModel's already-decoded string, even when that specific field was never touched.
- For entities whose raw bytes use character codes outside PKHeX's known tables (e.g. saves from fan-translation ROM hacks with custom character encodings not modeled by `StringConverter4`'s `TableINT`/`TableKOR`), decoding already produces an unrepresentable placeholder character. Re-encoding it replaces every unrecognized character with a literal `?`, permanently destroying the original bytes.
- `PokemonEditorViewModel.PreparePKM()` and `HallOfFame1EditorViewModel.SaveSlot()` now only re-encode when the value actually changed vs. the entity's currently decoded value, matching how `G4PKM.TransferTrashSwap` already avoids decode/re-encode for raw trash bytes during format conversion.
- Bumped `UIVersion` 1.51.4 → 1.51.5 (fix, patch bump).

## Test plan
- [x] `dotnet build PKHeX.sln -c Release` — 0 warnings, 0 errors
- [x] `dotnet test PKHeX.sln -c Release` — all suites pass (Core 449/450 +1 skipped, Architecture 5/5, Avalonia 2500/2500)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


doctorllll:
Opened by mistake due to a tool misconfiguration (gh defaulted to this upstream repo instead of the intended fork). Closing immediately — sorry for the noise.

